PR #3, opened 2020-09-11
Add Travis CI to the build

jeffmay:


project/Dependencies.scala
```diff
@@ -2,7 +2,7 @@ import sbt._
 
 object Dependencies {
 
-  final val Scala_2_13 = "2.13.3"
+  final val Scala_2_13 = "2.13.0"
```
jeffmay, line 5: This sucks, but I think it is related to https://github.com/typelevel/kind-projector/issues/116

PR #7, opened 2020-09-22
Better type selection and other cleanup

jeffmay:
- Use type selection builder at the top-level to avoid deprecated query
- Add Intersect and Union typeclasses for better or() / and() interface
- Define proper union typeclass for ResultSet
- Remove Query class and related methods
- Remove hasValue and filter dsl methods
- Replace ExpFunctor with the labeled ExpPure
- Allow use of infix "like" method for scalatest

### DIFF
--- a/core/src/main/scala/com/rallyhealth/vapors/core/data/NamedLens.scala
+++ b/core/src/main/scala/com/rallyhealth/vapors/core/data/NamedLens.scala
@@ -53,7 +53,8 @@ case class NamedLens[A, B](
 object NamedLens {
 
   type Id[A] = NamedLens[A, A]
-  def id[A]: NamedLens[A, A] = NamedLens(DataPath(Nil), identity[A])
+  val Id: NamedLens[Any, Any] = NamedLens(DataPath(Nil), identity[Any])
+  def id[A]: NamedLens[A, A] = Id.asInstanceOf[Id[A]]
 
   implicit class Selector[A, B](val lens: NamedLens[A, B]) extends AnyVal {
     def select[C](getter: B => C): NamedLens[A, C] = macro NamedLensMacros.selectImpl[A, B, C]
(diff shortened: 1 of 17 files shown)


core/src/main/scala/com/rallyhealth/vapors/core/data/NamedLens.scala
```diff
@@ -53,7 +53,8 @@ case class NamedLens[A, B](
 object NamedLens {
 
   type Id[A] = NamedLens[A, A]
-  def id[A]: NamedLens[A, A] = NamedLens(DataPath(Nil), identity[A])
+  val Id: NamedLens[Any, Any] = NamedLens(DataPath(Nil), identity[Any])
```
paul-a-kennedy-rally, line 56: should this be private?

core/src/main/scala/com/rallyhealth/vapors/core/data/NamedLens.scala
```diff
@@ -53,7 +53,8 @@ case class NamedLens[A, B](
...
+  def id[A]: NamedLens[A, A] = Id.asInstanceOf[Id[A]]
```
kionka-rally, line 57: I'm surprised this required an `asInstanceOf`. The types should be equivalent. Also, having something named `Id` vs `Id[A]` makes me a little nervous about shadowing.

core/src/main/scala/com/rallyhealth/vapors/core/data/Union.scala
```diff
@@ -0,0 +1,17 @@
...
+trait Union[A] {
+
+  def union(results: Seq[A]): A
+}
+
+object Union {
+
+  def apply[A](implicit A: Union[A]): Union[A] = A
+
+  implicit object boolean extends Union[Boolean] {
```
kionka-rally, line 12: These names should probably be longer since they are implicits. Something like `vaporsUnionBooleanInstance` or something.

paul-a-kennedy-rally (approved):

PR #9, opened 2020-09-27
Convert facts to be invariant

jeffmay:
- Added AbstractFact for better type inference
- Convert Fact, FactType, and ResultSet to be invariant
- Rename some type aliases to be more consistent with use of "Any"
  - Renamed AnyExp to Exp
  - Added AnyTerminalFactsExp for using abstract facts
- Moved the expression builders to common WhereExpBuilder interface


core/src/main/scala/com/rallyhealth/vapors/core/data/FactTypeSet.scala
```diff
@@ -1,71 +1,72 @@
...
-    if (typeMap.size != types.size) None
-    else Some(new FactTypeSet(typeMap))
+    NonEmptyMap
+      .fromMap(immutable.SortedMap.from(uniqueTypes))
+      .map { typeMap =>
+        Right(new FactTypeSet(typeMap))
+      }
+      .getOrElse {
+        Left(
+          s"types list contains duplicate fact types: ${duplicates.map(_._1).mkString("'", "', '", "'")}",
```
paul-a-kennedy-rally: We're not checking that duplicates is non-empty before creating a `Right()`.

This line will only be called if the list is empty (and not if there are duplicates).

core/src/main/scala/com/rallyhealth/vapors/core/data/NamedLens.scala
```diff
@@ -31,6 +34,42 @@ case class NamedLens[A, B](
     )
   }
 
+  // TODO: I don't think these make sense...
```
paul-a-kennedy-rally: remove?

core/src/main/scala/com/rallyhealth/vapors/core/data/ResultSet.scala
```diff
@@ -1,84 +1,122 @@
 package com.rallyhealth.vapors.core.data
 
-import cats.SemigroupK
 import cats.data.NonEmptyList
+import com.rallyhealth.vapors.core.dsl.{AbstractFacts, Facts}
 import com.rallyhealth.vapors.core.logic.{Intersect, Union}
 
-sealed trait ResultSet[+A] {
+sealed abstract class AbstractResultSet {
+  type FactValue
```
paul-a-kennedy-rally: Is this used?

core/src/main/scala/com/rallyhealth/vapors/core/data/ResultSet.scala
```diff
@@ -1,84 +1,122 @@
...
-
-object ResultSet {
-
-  def fromNel[I](nonEmptyFacts: NonEmptyList[Fact[I]]): FactsMatch[I] = FactsMatch(nonEmptyFacts)
+  @inline final def ++(o: AbstractResultSet): AbstractResultSet = this.union(o)
 
-  def fromList[I](facts: List[Fact[I]]): ResultSet[I] = NonEmptyList.fromList(facts) match {
-    case Some(nonEmptyList) => FactsMatch(nonEmptyList)
-    case None => NoFactsMatch()
+  override final def equals(o: Any): Boolean = o match {
```
paul-a-kennedy-rally: =, hashCode?

core/src/main/scala/com/rallyhealth/vapors/core/data/ResultSet.scala
```diff
@@ -1,84 +1,122 @@
...
+    _.reduceLeft[AbstractResultSet] {
       case (NoFactsMatch(), _) | (_, NoFactsMatch()) => NoFactsMatch() // skip the all expressions if the first failed
       case (acc, nextResult) => acc ++ nextResult // union all the required facts
     }
   }
 }
 
-final case class FactsMatch[+I](matchingFactsNel: NonEmptyList[Fact[I]]) extends ResultSet[I] {
+sealed trait ResultSet[A] extends AbstractResultSet {
+  override type FactValue = A
```
paul-a-kennedy-rally: Not needed?

core/src/main/scala/com/rallyhealth/vapors/core/data/ResultSet.scala
```diff
@@ -1,84 +1,122 @@
...
 
-  override def matchingFacts: List[Fact[I]] = matchingFactsNel.toList
+  override def matchingFacts: List[Fact[A]]
+}
+
+object ResultSet {
+
+  def fromNel[I](nonEmptyFacts: NonEmptyList[Fact[I]]): FactsMatch[I] = FactsMatch(nonEmptyFacts)
```
paul-a-kennedy-rally: Unify to use 'A' as the type parameter?

core/src/main/scala/com/rallyhealth/vapors/core/dsl/factfilter/Dsl.scala
```diff
@@ -40,34 +40,35 @@ private[dsl] class Dsl extends Types with Syntax {
...
 
-  def withFactsOfType[U : ClassTag : TypeTag](factType: FactType[U]): TypedFactExpBuilder[Any, U] = {
-    withFactsOfTypeIn[Any, U](FactTypeSet.of(factType))
+  def withFactsOfType[T >: U, U : ClassTag : TypeTag](factType: FactType[U]): WhereFactsExpBuilder[T, U] = {
+    new WhereFactsExpBuilder[T, U](FactTypeSet.of(factType))
   }
 
-  def withFactsOfTypeIn[T >: U, U : ClassTag : TypeTag](factTypeSet: FactTypeSet[U]): TypedFactExpBuilder[T, U] = {
-    new TypedFactExpBuilder[T, U](factTypeSet)
+  // TODO: Use shapeless for this
```
paul-a-kennedy-rally: Remove?  Or planning to do this?

core/src/main/scala/com/rallyhealth/vapors/core/dsl/factfilter/Dsl.scala
```diff
@@ -78,20 +79,21 @@ private[dsl] class Dsl extends Types with Syntax {
...
+  private def liftAnyTermExp(value: ExpAlg[AbstractFacts, AbstractResultSet]): AnyTerminalFactsExp =
+    FreeApplicative.lift(value)
 
   /** @see [[CondExp]] */
   private def liftCondExp[X](value: ExpAlg[X, Boolean]): CondExp[X] = FreeApplicative.lift(value)
 
-  /** @see [[AnyExp]] */
-  private def liftAnyExp[X, A](value: ExpAlg[X, A]): AnyExp[X, A] = FreeApplicative.lift(value)
+  /** @see [[Exp]] */
+  private def liftExp[X, A](value: ExpAlg[X, A]): Exp[X, A] = FreeApplicative.lift(value)
```
paul-a-kennedy-rally: value or "expression"

core/src/main/scala/com/rallyhealth/vapors/core/data/FactTypeSet.scala
jeffmay: Good catch! Fixed.

core/src/main/scala/com/rallyhealth/vapors/core/data/NamedLens.scala
jeffmay: Done.

core/src/main/scala/com/rallyhealth/vapors/core/data/ResultSet.scala
jeffmay: Nope! Removed.

core/src/main/scala/com/rallyhealth/vapors/core/data/ResultSet.scala
jeffmay: Nope. Removed.

core/src/main/scala/com/rallyhealth/vapors/core/dsl/factfilter/Dsl.scala
jeffmay: Good idea, I'll rename these to expression.

core/src/main/scala/com/rallyhealth/vapors/core/dsl/factfilter/Dsl.scala
jeffmay: I think I went with removing type parameters to avoid needing shapeless. Restricting input types makes the interface harder to use without much value. My gamble is that any expression at this level will require some form of type selection because relying on the Scala type is not specific enough to select the fact types you would want to filter to. For example, you might have "weight in lbs" represented in some Scala type, but that doesn't tell you where that measurement came from. You might want a fact representing "self-reported weight" and another representing "weight measurement at the doctor's office". Both are the same type of value, but they are different types of facts. Shapeless can't really help us here, so the payoff for using it throughout this level of the expression doesn't really buy us much.

There may be other use cases where shapeless would be a good fit in the future, but I don't think we need it yet.

core/src/main/scala/com/rallyhealth/vapors/core/data/ResultSet.scala
jeffmay: Good catch. Fixed.

core/src/main/scala/com/rallyhealth/vapors/core/data/ResultSet.scala
jeffmay: Good call. This is inconsistent with elsewhere. I've fixed it.

PR #10, opened 2020-09-29
Disable ScalaDoc generation to fix sbt publish

jeffmay:
- Disable ScalaDoc generation to fix publish
- Fix ScalaDoc link in Types trait

### DIFF
--- a/core/src/main/scala/com/rallyhealth/vapors/core/dsl/factfilter/Types.scala
+++ b/core/src/main/scala/com/rallyhealth/vapors/core/dsl/factfilter/Types.scala
@@ -40,7 +40,8 @@ private[dsl] trait Types {
   /**
     * An expression that terminates into a boolean, used for making a conditional query or filter.
     *
-    * Useful for defining sub-expressions to [[whereAnyFactHas]], [[whereAllFactsHave]], [[exists]], [[all]], etc.
+    * Useful for defining sub-expressions to [[WhereFactsExpBuilder.whereAllFacts]],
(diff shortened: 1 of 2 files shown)


kionka-rally (approved):

core/src/main/scala/com/rallyhealth/vapors/core/dsl/factfilter/Types.scala
```diff
@@ -40,7 +40,8 @@ private[dsl] trait Types {
   /**
     * An expression that terminates into a boolean, used for making a conditional query or filter.
     *
-    * Useful for defining sub-expressions to [[whereAnyFactHas]], [[whereAllFactsHave]], [[exists]], [[all]], etc.
+    * Useful for defining sub-expressions to [[WhereFactsExpBuilder.whereAllFacts]],
+    * [[WhereFactsExpBuilder.whereAnyValue]], [[exists]], [[all]], etc.
```
jeffmay, line 44: This change still didn't satisfy the ScalaDoc compiler, but it fixed it for IntelliJ IDEA.

darmbrus (approved):

paul-a-kennedy-rally:
Not everyone learns in the same way.

I do a lot better with hyperlinked documentation (and the various interfaces to it).

By not having buildable docs in the very first version, I am afraid that this will doom this project to never having buildable docs because the activation energy will be too high.

I have contributed documentation to projects where I've had to poke into something and figure it out as an outsider _and_ the docs are buildable.  Without buildable docs, you're back in the territory of having to use the IDE which exists for dealing with code at a different level.

I suspect that I'm not going to win this argument, but I wanted to register my objection.

PR #11, opened 2020-09-30
Move factfilter code out of core package

jeffmay:
- Move `dsl.factfilter` code to `factfilter.dsl` and `factfilter.data`
- Split fact data types out of `core.data` package into `factfilter.data`
- Moved fact type and object aliases to the `factfilter.data` package


paul-a-kennedy-rally (approved):
Looks good to me.

I'll probably have some fun refactoring the tests that I have.

PR #13, opened 2020-10-05
Add WindowSpec and fix a bug

jeffmay:
- Add unit tests for Window methods across multiple types
- Update scalafmt to allow min, min infix notation
- Reuse some Window methods
- Add scalatestplus-scalacheck and scalacheck-ops to the project
- Fix bug in BoundedWindow

### DIFF
--- a/core/src/main/scala/com/rallyhealth/vapors/core/data/Window.scala
+++ b/core/src/main/scala/com/rallyhealth/vapors/core/data/Window.scala
@@ -112,8 +112,8 @@ object Window {
       case Ior.Left(lb) if lb.inclusiveLowerBound => _ >= lb.lowerBound
       case Ior.Left(lb) => _ > lb.lowerBound
 
-      case Ior.Right(ub) if ub.inclusiveUpperBound => _ >= ub.upperBound
-      case Ior.Right(ub) => _ > ub.upperBound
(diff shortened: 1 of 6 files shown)


core/src/main/scala/com/rallyhealth/vapors/core/data/Window.scala
```diff
@@ -75,19 +75,18 @@ object Window {
     inclusive: Boolean,
   ): Window[A] = KnownWindow(Ior.Right(Below(max, inclusive)))
 
-  def lessThan[A : Ordering](max: A): Window[A] = KnownWindow(Ior.Right(Below(max, inclusiveUpperBound = false)))
+  @inline def lessThan[A : Ordering](max: A): Window[A] = lessThan(max, inclusive = false)
```
kionka-rally: I think we should remove these `inline`s. https://www.lightbend.com/blog/scala-inliner-optimizer

kionka-rally (approved):

paul-a-kennedy-rally (approved):
I like how complete this testing setup is.  I found it a little hard to unwind at first to understand how it was all being called.

I think that the code organization/layout isn't helping, but after spending a little time trying to shuffle it around, I'm not entirely sure that was going to help.

I'm not quite satisfied with how the tests are structured, but I can't offer anything concrete in terms of "I want this".  It's also possible that one needs to get more used to this form of testing.

PR #14, opened 2020-10-08
Fix use of __ outside of vapors library

jeffmay:
It turns out that __.withFactsOfType didn’t actually work outside of this library. This separates the use of __ for infix and top-level operators from the other more generic dsl functions.


paul-a-kennedy-rally (approved):
LGTM

kionka-rally (approved):
LGTM

PR #16, opened 2020-10-12
Better where expressions

jeffmay:
- Use a builder pattern for carrying type information that allows selecting from a single fact type multiple times
- Use a second builder pattern for operating on iterable collections with the same style


core/src/main/scala/com/rallyhealth/vapors/factfilter/dsl/Dsl.scala
```diff
@@ -9,14 +9,12 @@ import com.rallyhealth.vapors.factfilter.data._
 
 private[dsl] class Dsl extends TypedFactOps {
 
-  // TODO: Use some cats typeclass instead of iterable?
-  def all[F[x] <: IterableOnce[x], T, A](cond: CondExp[T]): CondExp[F[T]] = liftCondExp {
-    ExpAlg.ForAll[F[T], T, Boolean](identity[F[T]], cond, True, False)
+  def forall[T, V](cond: CondExp[V])(implicit ev: T <:< IterableOnce[V]): CondExp[T] = liftCondExp {
```
kionka-rally, line 12: Is `IterableOnce` a problem? "A template trait for collections which can be traversed either once only or one or more times." I think free applicatives can perform all the operations in one pass, but I don't know if this library has some parts which might want to go through twice.

kionka-rally (approved):

core/src/main/scala/com/rallyhealth/vapors/factfilter/dsl/Dsl.scala
jeffmay: This is because I want to use the early breaking "forall" and "exists" methods on the collections, (rather than computing all of the expressions in parallel and then ignoring the ones that are not needed to prove the query true). I'm not sure if `IterableOnce` will be the right solution, but it is a common supertype of `Option` and `Iterable`. I mainly only use this for facts that have nested data structures, so there isn't much concern about these not being in memory. Maybe someday I want to allow a fact to have a stream of data... but I can't see that for the forseeable future.

Long story short, this is safe for now, since I only traverse it once, but if I need more capabilities in the future, it shouldn't be hard to refactor because this is only used for data that has facts with non-scalar data fields.

core/src/main/scala/com/rallyhealth/vapors/factfilter/dsl/WhereExpBuilder.scala
```diff
@@ -1,55 +1,95 @@
...
+    new IterableOps[T, C, V](builder.wrap)
   }
 
-  def whereAllFacts(condExp: CondExp[TypedFact[U]]): TerminalFactsExp = {
-    new WhereValuesExpBuilder[T, U, TypedFact[U]](factTypeSet, NamedLens.id[TypedFact[U]]).whereAllValues(condExp)
+}
+
+final class CondBuilder[T, U](private val lens: NamedLens[T, U]) extends AnyVal {
+
+  private[CondBuilder] def wrap(cond: CondExp[U]): CondExp[T] =
```
jeffmay, line 42: I think this can just be `private`...

core/src/test/scala/com/rallyhealth/vapors/core/fql/PrinterSpec.scala
```diff
@@ -25,10 +25,10 @@ class PrinterSpec extends AnyWordSpec {
       }
     }
 
-    "print an expression that selects a fact type using where(factValue(_ > 200))" in {
+    "print an expression that selects a fact type using whereAnyFact(_.at(_.select(_.value)) > 200))" in {
       val q = {
-        __.withFactsOfType(FactTypes.WeightMeasurement)
-          .whereAnyValue(__ > 200)
+        __.withType(FactTypes.WeightMeasurement)
```
jeffmay: With single underscore `_` now in use because of the `ConditionOps`, I'm not sure if I like the double underscore as much. Originally, it was meant to be a way to access all of the operations using auto-complete (including infix operators), but it isn't even doing that anymore (it only does the old-style infix operators with type selector methods).

Either I should go back to making `__` an alias to the full DSL (with or without infix operators), or I should just remove it and let the `dsl` package object be the starting point for the full DSL.

core/src/test/scala/com/rallyhealth/vapors/core/fql/PrinterSpec.scala
```diff
@@ -43,27 +43,25 @@ class PrinterSpec extends AnyWordSpec {
...
       assertResult(FactsMatch(NonEmptyList.of(JoeSchmoe.weight))) {
         evalWithFacts(JoeSchmoe.facts)(q)
       }
       assertResult(
-        // TODO: This query looks wrong...
-        //       there is a mismatch between how value selection works within .exists() and conditional expressions
         "_.collect {" +
           " case f: Fact[Example.Probs] =>" +
-          " f.exists(_.value.scores['weightloss'] .exists(_where x > 0.5) and .exists(_where x < 0.3)) " +
+          " f.exists(_.value.scores['weightloss'] .exists(_ where x > 0.5 and  where x < 0.3)) " +
```
paul-a-kennedy-rally, line 62: The expression seems to have gone from `_where` in both cases previously to something inconsistent now?

Is this a bug in a pretty printer?

jeffmay: This code is unused and I plan to remove it anyway.

jeffmay: See https://github.com/jeffmay/vapors/pull/15

PR #17, opened 2020-10-15
Add README.md

jeffmay:


paul-a-kennedy-rally (approved):

README.md
```diff
@@ -1 +1,253 @@
-# vapor
\ No newline at end of file
+![Vapor Logo](docs/logo.png)
+
+# Vapors
+
+The Vapors library provides an embedded-DSL for writing logical queries over a collection of facts.
+It is based on the `cats-free` implementation of `FreeApplicative` for building the underlying algebra
```
kionka-rally: We might want to move these descriptions to the implementation section.

README.md
```diff
@@ -1 +1,253 @@
...
+2.  **Define your fact types.**
+
+    Every fact type must have a unique name as well as a Scala type. You can define a FactType with any Scala type you
+    want, but if you want to be able to use comparison operators, you must define an `Ordering`. To check equality, you
+    need to define `cats.Eq`.
+
+    ```
+    object FactTypes {
+      val DateOfBirth = FactType[LocalDate]("DateOfBirth")
+      val Role = FactType[String]("Role")
```
kionka-rally: `String` should be `Role`. Also, I think you might need to rename `Role` because of shadowing. Also, the code below needs to use `Role`.

README.md
```diff
@@ -1 +1,253 @@
...
+
+    ```
+    resolvers += Resolver.bintray("jeffmay", "maven")
+    libraryDependencies += "com.rallyhealth" %% "vapors" % "0.1.0"
+    ```
+
+2.  **Define your fact types.**
+
+    Every fact type must have a unique name as well as a Scala type. You can define a FactType with any Scala type you
+    want, but if you want to be able to use comparison operators, you must define an `Ordering`. To check equality, you
```
kionka-rally, line 48: There's a Cats `Order` that we might as well use if we are requiring the Cats `Eq`.

README.md
```diff
@@ -1 +1,253 @@
...
+    You must start with either a `TerminalFactsExp` or a logical operator (such as `and` / `or`). You will typically build a
+    `TerminalFactsExp` by filtering to a specific type of fact and applying a filter condition on the instances. Since every
+    `FactType` can have multiple instances, you have to decide how to filter the list with a logical rule:
+    
+    ```
+    import cats.implicits._
+    import com.rallyhealth.vapors.factfilter.dsl._
+    
+    val isAdmin = {
+      withType(FactTypes.Role).whereAnyFactValue(_ === "admin")
```
kionka-rally: I understand why this is done in two steps, but I still wonder if there's another way.

README.md
```diff
@@ -1 +1,253 @@
...
+    libraryDependencies += "com.rallyhealth" %% "vapors" % "0.1.0"
+    ```
+
+2.  **Define your fact types.**
+
+    Every fact type must have a unique name as well as a Scala type. You can define a FactType with any Scala type you
+    want, but if you want to be able to use comparison operators, you must define an `Ordering`. To check equality, you
+    need to define `cats.Eq`.
+
+    ```
```
kionka-rally: Needs `scala` here and below.

README.md
```diff
@@ -1 +1,253 @@
...
+In essence, every query can be optimized and evaluated to produce a function: 
+```
+Facts => ResultSet
+```
+
+### Types of Expressions
+
+| Type | Description | Definition |
+| ---- | ----------- | ---------- |
+| Exp  | A generic expression. | `FreeApplicative[ExpAlg[T, *, ...], A]`* |
```
kionka-rally: `FreeApplicative[ExpAlg[T, *, ...], A]`* displays weird and needs more explanation probably.

README.md
```diff
@@ -1 +1,253 @@
...
+val rs = evalWithFacts(JoeSchmoe.facts)(query)
+assert(rs.isTrue)
+assert(rs.matchingFacts == Facts(JoeSchmoe.height, JoeSchmoe.weight, JoeSchmoe.predictWeightloss))
+```
+
+#### DSL Operators
+
+| Operator | Sub-Expression Type | Return Type | Description |
+| -------- | ------------------- | ----------- | ----------- |
+| `and[T, A]` / `&&[T, A]` | `Exp[T, A]` | Builds an expression that returns a falsey `A` when any sub-expression returns a falsey `A` result |
```
kionka-rally: This description doesn't show up.

README.md
```diff
@@ -1 +1,253 @@
...
+| `or[T, A]` / `||[T, A]` | `Exp[T, A]` | Builds an expression that returns a truthy `A` when any sub-expression returns a truthy `A` result |
```
kionka-rally: Display errors on this line too.

README.md
```diff
@@ -1 +1,253 @@
...
+In order to interpret expressions built by our embedded DSL, we wrap our `ExpAlg` algebraic data type (ADT)
+in a `FreeApplicative` data structure from the `cats-free` library. This allows us to use operations such
+as `.map` and `.foldMap` to produce new applicative contexts that can be evaluated efficiently.
+
+### Interpreters
+
+In order to go from `Exp[T, A]` to another type, such as a Json `Encoder[A]` or `Decoder[A]`, you just need
+to define a *natural transformation* (also known as a `FunctionK[F[_], G[_]]`) from `ExpAlg[T, *] ~> G`.
+
+Check out these [helpful slides on the library design](
```
kionka-rally, line 400: I think some more description can go here. Maybe some links to free applicative information here.

kionka-rally (commented):
Talked about Jeff with this for a while IRL. Might also be worth talking about how facts are not "true" or "false", they are "true" or cannot be proven "true" with the given facts. The interesting aspect of the library is how to returns the matching facts. https://en.wikipedia.org/wiki/Intuitionistic_logic

README.md
jeffmay: I'm actually thinking about removing `Eq` and just using Scala's `Ordering` for everything.

PR #19, opened 2020-10-26
Add between method

jeffmay:


core/src/main/scala/com/rallyhealth/vapors/factfilter/dsl/Dsl.scala
```diff
@@ -17,20 +17,8 @@ private[dsl] class Dsl extends TypedFactOps {
...
-  }
-
-  def greaterThan[T : Ordering](lowerBound: T): CondExp[T] = liftCondExp {
-    ExpAlg.Within[T, Boolean](Window.greaterThan(lowerBound), True, False)
-  }
-
-  def greaterThanOrEqual[T : Ordering](lowerBound: T): CondExp[T] = liftCondExp {
-    ExpAlg.Within[T, Boolean](Window.greaterThanOrEqual(lowerBound), True, False)
+  def within[T](window: Window[T]): CondExp[T] = liftCondExp {
+    ExpAlg.Within[T, Boolean](window, True, False)
```
paul-a-kennedy-rally, line 21: OK, we have some serious case of boolean blindness here and it is too easy to make mistakes.

Maybe add:
* `excludeLowerBound`
* `includeLowerBound`
* `excludeUpperBound`
* `includeUpperBound`

paul-a-kennedy-rally (commented):
I like the ideas, but just from this review (and the tests), I cannot tell what the bounds are and how this work without having to dig into the larger code.

core/src/main/scala/com/rallyhealth/vapors/factfilter/dsl/Dsl.scala
jeffmay: These cases are all covered by the WindowSpec.

paul-a-kennedy-rally: I suppose what I'm getting at is that by looking at this expression: 
`expAlg.Within[T, Boolean](window, True, False)` I don't know what the `True` and `False` represent because this single bit of information has been stripped from its context.

I believe that I reviewed the PR that included the WindowSpec, and yes, it does test all the cases, but it wasn't without error initially (as I recall).

What I'm saying is have types so that instead of saying true/false,  you can put something with more context.

paul-a-kennedy-rally: Ok, I see how this is working now....  never mind :)

paul-a-kennedy-rally (approved):

PR #21, opened 2020-10-27
Limit results of where builders to only matching facts

jeffmay:
This fixes the bug where you have an expression that filters to a specific type like the following:
```
withType(Tag).whereAnyFactValue { tag =>
  tag === "A" or tag === "B"
}
```

And you have some facts of the same type that don't match the expression's filter:
```
Facts(
  Tag("A"),
  Tag("C"),
)
```

And the result used to return all facts of the selected type if at least one of them matched:
```
ResultSet(
  Facts(
    Tag("A"),
    Tag("C"),
  ),
)
```

But now it will only return the specific facts that matched:
```
ResultSet(
  Facts(
    Tag("A"),
  ),
)
```


paul-a-kennedy-rally (approved):
LGTM

PR #23, opened 2020-12-09
Tagless final encoding with new algebra for embedding expressions and processing results

jeffmay:
- Add ExprResult tree with Visitor and documentation
- Add Define and UsingDefinitions expression nodes to new algebra
- Add LogicalConnectives and LogicalExprSpec
- Add better toString for Fact and FactType
- Add timestamp to measurements for testing
- Add example of CaptureP called CaptureTimeRange
- Add EmbededExpressionSpec for testing an example embedded expression
- Make EmbeddedExpressionSpec pendingUntilFixed for derived facts
- Implement Negation, Conjunction, Disjunction
- Implement Negative, Addition, Subtraction
- Remove UnknownWindow
- Use Order instead of Ordering
- Add ExprBuilder pattern for new algebra

PR #25, opened 2020-12-15
Use simpler Evidence class and switch to SortedSet for facts

jeffmay:
- Create new Evidence class
- Use new simpler Evidence class
- Switch to SortedSet for facts in Evidence and FactTable

Depends on:
- [x] https://github.com/jeffmay/vapors/pull/23

### DIFF
--- a/core/src/main/scala/com/rallyhealth/vapors/factfilter/data/Evidence.scala
+++ b/core/src/main/scala/com/rallyhealth/vapors/factfilter/data/Evidence.scala
@@ -1,0 +1,71 @@
+package com.rallyhealth.vapors.factfilter.data
+
+import cats.Monoid
+import cats.data.NonEmptySet
+
+import scala.collection.immutable.SortedSet
+
+final class Evidence private (val factSet: SortedSet[Fact]) extends AnyVal {
+
+  def isEmpty: Boolean = factSet.isEmpty
+  def nonEmpty: Boolean = factSet.nonEmpty
+
+  @deprecated("Use factSet instead.", "0.3.0")
+  def facts: List[Fact] = factSet.toList
+
+  @inline def ++(that: Evidence): Evidence = union(that)
+  @inline def |(that: Evidence): Evidence = union(that)
+
+  def ofType[T](factTypeSet: FactTypeSet[T]): Option[NonEmptySet[TypedFact[T]]] = {
+    val matchingFacts = SortedSet.from(this.factSet.iterator.collect(factTypeSet.collector))
+    NonEmptySet.fromSet(matchingFacts)
+  }
+
+  /**
+    * Union of the two sets of results.
+    */
+  def union(that: Evidence): Evidence = this.factSet match {
+    case Evidence.none.factSet | that.factSet => that
+    case _ => Evidence(this.factSet | that.factSet)
+  }
+
+  override def toString: String =
+    if (this.factSet.isEmpty) "Evidence()"
+    else s"Evidence${factSet.mkString("(", ", ", ")")}"
+}
+
+object Evidence {
+
+  def unapply(evidence: Evidence): Some[SortedSet[Fact]] = Some(evidence.factSet)
+
+  final class SetOfFacts private[Evidence] (private[Evidence] val factSet: SortedSet[Fact]) extends AnyVal
+
+  implicit def singleSetOfFacts(fact: Fact): SetOfFacts = new SetOfFacts(SortedSet(fact))
+  implicit def iterableSetOfFacts(facts: Iterable[Fact]): SetOfFacts = new SetOfFacts(SortedSet.from(facts))
+
+  @inline final def apply(factSets: SetOfFacts*): Evidence = {
+    if (factSets.isEmpty) none
+    else new Evidence(factSets.iterator.foldLeft(none.factSet)(_ | _.factSet))
+  }
+
+  final val none = new Evidence(SortedSet.empty[Fact])
+
+  /**
+    * Unions evidence as a standard definition for monoid.
+    *
+    * This is the right behavior to satify [[Monoid]], however if you need to do evidence tracking,
+    * you wlil probably want to be more specific if you need to represent the behavior of merging
+    * evidence from the output of various expressions (or their input).
+    *
+    * @see [[com.rallyhealth.vapors.factfilter.evaluator.InterpretExprAsFunction.Output.monoid]]
+    */
+  implicit val monoid: Monoid[Evidence] = {
+    new Monoid[Evidence] {
+      override def empty: Evidence = Evidence.none
+      override def combine(
+        x: Evidence,
+        y: Evidence,
+      ): Evidence = x.union(y)
+    }
+  }
+}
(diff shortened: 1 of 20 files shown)


core/src/main/scala/com/rallyhealth/vapors/factfilter/data/Fact.scala
```diff
@@ -20,6 +22,36 @@ object Fact {
...
+  }
+
+  def orderByFactName(nameOrder: Order[String]): Order[Fact] = {
+    nameOrder.contramap(_.typeInfo.name)
+  }
+
+  // This will be used a lot, so cache it
+  private final val defaultOrder: Order[Fact] = orderByFactName(Order[String])
+
+  implicit def order(implicit orderFactNames: Order[String]): Order[Fact] = {
```
kionka-rally, line 51: So, these are mostly compared as `String`s?

core/src/main/scala/com/rallyhealth/vapors/factfilter/data/FactTable.scala
```diff
@@ -1,56 +1,70 @@
...
   }
 
-  def get[T](factType: FactType[T]): List[TypedFact[T]] =
-    facts.get(factType.fullName).toList.flatMap(_.collect(Function.unlift(factType.cast)))
-
-  def getAll[T](factTypeSet: FactTypeSet[T]): List[TypedFact[T]] = {
-    factTypeSet.typeMap.toSortedMap.values.foldLeft(List.empty[TypedFact[T]]) {
-      case (acc, ft) =>
-        acc ::: get(ft)
+  def getAllByFactType[T](factTypeSet: FactTypeSet[T]): TypedFactSet[T] = {
```
kionka-rally, line 24: `FactTypeSet` vs `TypedFactSet` is going to be very confusing.

core/src/main/scala/com/rallyhealth/vapors/factfilter/data/FactTypeSet.scala
```diff
@@ -2,34 +2,32 @@ package com.rallyhealth.vapors.factfilter.data
...
     */
-  def cast(fact: Fact): Option[TypedFact[A]] = castF(fact)
+  def cast(fact: Fact): Option[TypedFact[A]] = collector.lift(fact)
 
   /**
     *
     * @return a partial function for collecting facts of a specific type
     */
-  def collector: PartialFunction[Fact, TypedFact[A]] = castPF
+  def collector: PartialFunction[Fact, TypedFact[A]] = {
```
kionka-rally, line 33: There are... several issues with this function. Maybe there should be a requirement that there are enums of Facts provided?

Does this work by type inference? I guess by the type parameter of the class? Is that reliable?

core/src/main/scala/com/rallyhealth/vapors/factfilter/dsl/ExprBuilder.scala
```diff
@@ -79,6 +82,23 @@ final class SubtractBuilderOps[R : Subtraction](number: R) {
...
   extends ExprBuilder[F, V, M, U, P](returnOutput) {
 
+  def toList(
+    implicit
+    ev: M[U] <:< Iterable[U],
+    captureOutput: CaptureP[F, V, List[U], P],
+  ): FoldableExprBuilder[F, V, List, U, P] =
+    to(List)
+
+  def to[N[+_] <: Iterable[_] : Foldable](
```
kionka-rally: This is quite a type. Does it need to be Iterable and Foldable?

core/src/main/scala/com/rallyhealth/vapors/factfilter/dsl/ExprDsl.scala
```diff
@@ -10,17 +10,19 @@ import com.rallyhealth.vapors.core.math.{Addition, Negative, Subtraction}
 import com.rallyhealth.vapors.factfilter.data._
 import com.rallyhealth.vapors.factfilter.evaluator.InterpretExprAsFunction
 
+import scala.collection.immutable.SortedSet
+
 object ExprDsl extends ExprBuilderSyntax {
 
-  type CondExpr[F[_], V, P] = Expr[F, V, Boolean, P]
+  final type CondExpr[F[_], V, P] = Expr[F, V, Boolean, P]
```
kionka-rally, line 17: Does this being an object not already make these final? Should the rest of these be final? It's a shame that final is not the default in Scala.

core/src/main/scala/com/rallyhealth/vapors/factfilter/data/Fact.scala
jeffmay: yea, it is mostly a wrapper around a String with some type information.

core/src/main/scala/com/rallyhealth/vapors/factfilter/data/FactTable.scala
jeffmay: Yea, one is a `Set` of `FactType`s and the other is a `FactSet` with a specific type.

core/src/main/scala/com/rallyhealth/vapors/factfilter/data/FactTypeSet.scala
jeffmay: I am thinking of maybe using Shapeless's `TypeCase` instead of TypeTag, but it doesn't seem necessary at this point (maybe for Scala 3 cross-compatibility)

core/src/main/scala/com/rallyhealth/vapors/factfilter/dsl/ExprBuilder.scala
jeffmay: Good catch. I only need the `Foldable` typeclass.

core/src/main/scala/com/rallyhealth/vapors/factfilter/dsl/ExprDsl.scala
jeffmay: I'm not sure... I just put `final` everywhere because, as you say, it should be the default. I believe Scala 3 uses a new `open` keyword for making a class open for extension, and in Scala >3.1, it will be required to extend a class and all non-open classes will be automatically made `final`.

core/src/main/scala/com/rallyhealth/vapors/factfilter/data/Evidence.scala
```diff
@@ -0,0 +1,73 @@
...
+import cats.{Monoid, Order}
+import cats.data.NonEmptySet
+import cats.instances.order._
+
+import scala.collection.immutable.SortedSet
+
+final class Evidence private (val factSet: SortedSet[Fact]) extends AnyVal {
+
+  def isEmpty: Boolean = factSet.isEmpty
+  def nonEmpty: Boolean = factSet.nonEmpty
```
kionka-rally, line 12: I'm generally against any negatively phrased boolean concepts, since they nearly always immediately lead to double negation, and no humans are not bad at understanding that.

jeffmay: I thought the point of a `nonEmpty` method was to avoid a double negative at the call-site... So you can write `if (x.nonEmpty)` versus `if (!x.isEmpty)`

core/src/main/scala/com/rallyhealth/vapors/factfilter/data/FactTable.scala
paul-a-kennedy-rally: Is there some documentation (usage?) to which someone could refer?

paul-a-kennedy-rally (approved):
No blockers, but I do worry about some of the class names not being clear enough.

core/src/main/scala/com/rallyhealth/vapors/factfilter/data/FactTable.scala
jeffmay: I'll update this

kionka-rally (approved):

PR #26, opened 2020-12-16
Add support for derived facts

jeffmay:
- Adds support for facts that are derived from other fact expression results

Depends on:
- [x] https://github.com/jeffmay/vapors/pull/25

### DIFF
--- a/core/src/main/scala/com/rallyhealth/vapors/factfilter/data/Evidence.scala
+++ b/core/src/main/scala/com/rallyhealth/vapors/factfilter/data/Evidence.scala
@@ -1,8 +1,8 @@
 package com.rallyhealth.vapors.factfilter.data
 
-import cats.{Monoid, Order}
 import cats.data.NonEmptySet
 import cats.instances.order._
+import cats.{Monoid, Order}
 
 import scala.annotation.tailrec
 import scala.collection.immutable.SortedSet
@@ -66,6 +66,28 @@ object Evidence {
   }
 
   final val none = new Evidence(SortedSet.empty[Fact])
+
+  /**
+    * Convert any given value into [[Evidence]] by inspecting whether it is a fact or valid collection of facts.
+    *
+    * This is used by the library when iterating over a collection of facts, where the facts can be used as their own
+    * evidence in the subexpression.
+    */
+  def fromAny(any: Any): Option[Evidence] = any match {
+    case ev: Evidence => Some(ev)
+    case fact: Fact => Some(Evidence(fact))
+    case map: collection.Map[_, _] => fromAnyIterable(map.valuesIterator)
+    case iter: IterableOnce[_] => fromAnyIterable(iter)
+    case _ => None
+  }
+
+  @inline def fromAnyOrNone(any: Any): Evidence = fromAny(any).getOrElse(none)
+
+  private[this] def fromAnyIterable(anyIter: IterableOnce[_]): Option[Evidence] = {
+    val iter = anyIter.iterator
+    if (iter.isEmpty) None
+    else Some(Evidence(FactSet.from(iter.collect { case fact: Fact => fact })))
+  }
 
   /**
     * Unions evidence as a standard definition for monoid.
--- a/core/src/main/scala/com/rallyhealth/vapors/factfilter/data/Evidence.scala
+++ b/core/src/main/scala/com/rallyhealth/vapors/factfilter/data/Evidence.scala
@@ -32,7 +32,6 @@ final class Evidence private (val factSet: SortedSet[Fact]) extends AnyVal {
     case _ => Evidence(this.factSet | that.factSet)
   }
 
-  // TODO: Use a separate type?
   def derivedFromSources: Evidence = {
     @tailrec def loop(
       mixed: Iterable[Fact],
--- a/core/src/main/scala/com/rallyhealth/vapors/factfilter/data/Evidence.scala
+++ b/core/src/main/scala/com/rallyhealth/vapors/factfilter/data/Evidence.scala
@@ -4,6 +4,7 @@ import cats.{Monoid, Order}
 import cats.data.NonEmptySet
 import cats.instances.order._
 
+import scala.annotation.tailrec
 import scala.collection.immutable.SortedSet
 
 final class Evidence private (val factSet: SortedSet[Fact]) extends AnyVal {
@@ -31,6 +32,25 @@ final class Evidence private (val factSet: SortedSet[Fact]) extends AnyVal {
     case _ => Evidence(this.factSet | that.factSet)
   }
 
+  // TODO: Use a separate type?
+  def derivedFromSources: Evidence = {
+    @tailrec def loop(
+      mixed: Iterable[Fact],
+      source: FactSet,
+    ): FactSet = {
+      if (mixed.isEmpty) source
+      else {
+        val (remainingEvidence, sourceFacts) = mixed.partitionMap {
+          case DerivedFact(_, _, evidence) => Left(evidence.factSet)
+          case source => Right(source)
+        }
+        loop(remainingEvidence.flatten, source ++ sourceFacts)
+      }
+    }
+    val allSourceFacts = loop(this.factSet, FactSet.empty)
+    new Evidence(allSourceFacts)
+  }
+
   override def toString: String =
     if (this.factSet.isEmpty) "Evidence()"
     else s"Evidence${factSet.mkString("(", ", ", ")")}"
@@ -40,14 +60,9 @@ object Evidence {
 
   def unapply(evidence: Evidence): Some[SortedSet[Fact]] = Some(evidence.factSet)
 
-  final class SetOfFacts private[Evidence] (private[Evidence] val factSet: SortedSet[Fact]) extends AnyVal
-
-  implicit def singleSetOfFacts(fact: Fact): SetOfFacts = new SetOfFacts(SortedSet(fact))
-  implicit def iterableSetOfFacts(facts: Iterable[Fact]): SetOfFacts = new SetOfFacts(SortedSet.from(facts))
-
-  @inline final def apply(factSets: SetOfFacts*): Evidence = {
-    if (factSets.isEmpty) none
-    else new Evidence(factSets.iterator.foldLeft(none.factSet)(_ | _.factSet))
+  @inline final def apply(facts: FactOrFactSet*): Evidence = {
+    if (facts.isEmpty) none
+    else new Evidence(FactOrFactSet.flatten(facts))
   }
 
   final val none = new Evidence(SortedSet.empty[Fact])
--- a/core/src/main/scala/com/rallyhealth/vapors/factfilter/data/Fact.scala
+++ b/core/src/main/scala/com/rallyhealth/vapors/factfilter/data/Fact.scala
@@ -18,12 +18,6 @@ object Fact {
     value: T,
   ): Fact = SourceFactOfType(factType, value)
 
-  def apply[T](
-    factType: FactType[T],
-    value: T,
-    evidence: Evidence,
-  ): DerivedFact = DerivedFactOfType(factType, value, evidence)
-
   def unapply(fact: Fact): Some[(FactType[fact.Value], fact.Value)] = Some((fact.typeInfo, fact.value))
 
   val orderByFactValue: Order[Fact] = { (x, y) =>
--- a/core/src/main/scala/com/rallyhealth/vapors/factfilter/data/Fact.scala
+++ b/core/src/main/scala/com/rallyhealth/vapors/factfilter/data/Fact.scala
@@ -9,19 +9,22 @@ sealed abstract class Fact {
 
   val typeInfo: FactType[Value]
   val value: Value
-
-  override def toString: String = {
-    // TODO: JSON format for value? Show for value?
-    s"Fact(${typeInfo.fullName} = $value)"
-  }
 }
 
 object Fact {
 
-  @inline final def apply[T](
+  def apply[T](
     factType: FactType[T],
     value: T,
-  ): Fact = TypedFact(factType, value)
+  ): Fact = SourceFactOfType(factType, value)
+
+  def apply[T](
+    factType: FactType[T],
+    value: T,
+    evidence: Evidence,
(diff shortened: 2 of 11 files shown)


core/src/main/scala/com/rallyhealth/vapors/factfilter/data/Fact.scala
```diff
@@ -10,18 +10,29 @@ sealed abstract class Fact {
   val typeInfo: FactType[Value]
   val value: Value
 
-  override def toString: String = {
-    // TODO: JSON format for value? Show for value?
-    s"Fact(${typeInfo.fullName} = $value)"
+  // TODO: JSON format for value? Show for value?
```
kionka-rally: `Show` should have been called `Debug` if that makes sense. Another typeclass should be used for JSON representations.

core/src/main/scala/com/rallyhealth/vapors/factfilter/data/Fact.scala
```diff
@@ -10,18 +10,29 @@ sealed abstract class Fact {
...
+  override def toString: String = this match {
+    case derivedFact: DerivedFact =>
```
kionka-rally: Why are you using reflection here?

core/src/main/scala/com/rallyhealth/vapors/factfilter/data/Fact.scala
```diff
@@ -54,15 +65,28 @@ object Fact {
...
 }
 
 object TypedFact {
 
+  final def apply[A](
+    typeInfo: FactType[A],
+    value: A,
+  ): TypedFact[A] = SourceFactOfType(typeInfo, value)
+
+  final def apply[A](
```
kionka-rally: I'm not a big fan of the method overloading here.

core/src/main/scala/com/rallyhealth/vapors/factfilter/data/Fact.scala
```diff
@@ -80,3 +104,39 @@ object TypedFact {
...
+    case _ => None
+  }
+}
+
+final case class SourceFactOfType[A](
+  typeInfo: FactType[A],
+  value: A,
+) extends TypedFact[A]
+
+object SourceFactOfType {}
```
kionka-rally: What is the purpose of the empty companion objects?

core/src/main/scala/com/rallyhealth/vapors/factfilter/data/Fact.scala
```diff
@@ -54,15 +65,28 @@ object Fact {
...
+
+  final def apply[A](
+    typeInfo: FactType[A],
+    value: A,
+    evidence: Evidence,
+  ): TypedFact[A] with DerivedFact = DerivedFactOfType(typeInfo, value, evidence)
+
+  final def unapply[A](fact: TypedFact[A]): Some[(FactType[A], A)] = fact match {
+    case SourceFactOfType(typeInfo, value) => Some((typeInfo, value))
+    case DerivedFactOfType(typeInfo, value, _) => Some((typeInfo, value))
```
kionka-rally, line 73: Does this lose important information?

core/src/main/scala/com/rallyhealth/vapors/factfilter/data/FactOrFactSet.scala
```diff
@@ -0,0 +1,20 @@
+package com.rallyhealth.vapors.factfilter.data
+
+import cats.Order
+import cats.instances.order._
+
+import scala.collection.immutable.SortedSet
+
+final class FactOrFactSet private[FactOrFactSet] (val toSortedSet: SortedSet[Fact]) extends AnyVal
```
kionka-rally, line 8: `FactOrFactSet` is probably asking for another name. Perhaps just `FactSet`?

core/src/main/scala/com/rallyhealth/vapors/factfilter/data/FactOrFactSet.scala
```diff
@@ -0,0 +1,20 @@
...
+
+object FactOrFactSet {
+
+  implicit def setOfOneFact(fact: Fact)(implicit order: Order[Fact]): FactOrFactSet = new FactOrFactSet(SortedSet(fact))
```
kionka-rally, line 12: These implicits seem a bit aggressive.

core/src/main/scala/com/rallyhealth/vapors/factfilter/data/Fact.scala
jeffmay: Good point. It's a `sealed trait`, so I can just override toString in the case classes.

core/src/main/scala/com/rallyhealth/vapors/factfilter/evaluator/InterpretExprAsFunction.scala
```diff
@@ -377,7 +393,10 @@ object InterpretExprAsFunction {
   final case class Output[R](
     value: R,
     evidence: Evidence,
-  )
+  ) {
+
+    def addEvidence(evidence: Evidence): Output[R] = copy(evidence = this.evidence ++ evidence)
```
jeffmay: Is this still used?

core/src/main/scala/com/rallyhealth/vapors/factfilter/evaluator/InterpretExprAsFunction.scala
```diff
@@ -252,18 +263,20 @@ final class InterpretExprAsFunction[F[_] : Foldable, V, P]
...
-    val postParam = expr.capture.foldToParam(expr, input, subResult.output, allParams)
-    ExprResult.UsingDefinitions(expr, ExprResult.Context(input, subResult.output, postParam), subResult)
+    val postParam = expr.capture.foldToParam(expr, input, output, allParams)
+    ExprResult.UsingDefinitions(expr, ExprResult.Context(input, output, postParam), subResult)
   }
 
   override def visitWhen[R](expr: Expr.When[F, V, R, P]): Input[F, V] => ExprResult[F, V, R, P] = { input =>
     val conditionResult = expr.condExpr.visit(this)(input)
+    // TODO: Should this include the original input.evidence?
     val subInput = input.copy(evidence = conditionResult.output.evidence)
```
jeffmay: Probably not...

core/src/main/scala/com/rallyhealth/vapors/factfilter/evaluator/InterpretExprAsFunction.scala
```diff
@@ -213,7 +224,7 @@ final class InterpretExprAsFunction[F[_] : Foldable, V, P]
   }
 
   override def visitReturnInput(expr: Expr.ReturnInput[F, V, P]): Input[F, V] => ExprResult[F, V, F[V], P] = { input =>
-    resultOfPureExpr(expr, input, input.value, input.evidence) {
+    resultOfPureExpr(expr, input, input.value, input.evidence ++ Evidence.fromAnyOrNone(input.value)) {
```
jeffmay, line 227: I should document this. Basically, if the input is a Fact or an Iterable[Fact], then it is added to its own evidence when moved to the output value.

core/src/main/scala/com/rallyhealth/vapors/factfilter/evaluator/InterpretExprAsFunction.scala
```diff
@@ -128,7 +135,9 @@ final class InterpretExprAsFunction[F[_] : Foldable, V, P]
     val inputResult = expr.inputExpr.visit(this)(input)
     val flatMapFn = expr.flatMapExpr.visit(InterpretExprAsFunction())
     val allResults = inputResult.output.value.map { elem =>
-      val flatMapInput = input.withValue(elem)
+      // If given a fact, use it as evidence, otherwise keep input evidence for this value of the collection
+      val inputEvidence = Evidence.fromAny(elem).getOrElse(inputResult.output.evidence)
```
jeffmay, line 139: The comment is repeated 3 times... I should probably move it to helper code.

core/src/main/scala/com/rallyhealth/vapors/factfilter/data/Evidence.scala
```diff
@@ -31,6 +32,25 @@ final class Evidence private (val factSet: SortedSet[Fact]) extends AnyVal {
     case _ => Evidence(this.factSet | that.factSet)
   }
 
+  // TODO: Use a separate type?
```
jeffmay: Remove this TODO. I don't think it is helpful to have a separate type.

core/src/main/scala/com/rallyhealth/vapors/factfilter/data/FactOrFactSet.scala
jeffmay: There already is a `FactSet`. This is magnet type. Its only purpose is to avoid method overloading for the `Evidence` and `FactTable` factory methods.

core/src/main/scala/com/rallyhealth/vapors/factfilter/data/Fact.scala
jeffmay: If the evidence field were important, the caller would use the `DerivedFact` extractor object.

core/src/main/scala/com/rallyhealth/vapors/factfilter/data/FactOrFactSet.scala
jeffmay: This is the main purpose for this `FactOrFactSet` type. It is the magnet pattern to avoid overloading the factory methods for `FactTable` and `Evidence`. It's syntactic sugar for consistent wrapping facts (or sets of facts) as FactTables or Evidence.

core/src/main/scala/com/rallyhealth/vapors/factfilter/dsl/ExprBuilder.scala
```diff
@@ -37,10 +37,11 @@ object ExprBuilder extends ExprBuilderSyntax {
 
 trait ExprBuilderLowPriorityImplicits {
 
+  // TODO: Add unit test for this
   implicit def liftFoldableExpr[F[_] : Foldable, V, M[_] : Foldable, U, P](
```
jeffmay: I don't think this is used, and I'm not sure it even works. I should write a unit test to verify or remove this code.

core/src/main/scala/com/rallyhealth/vapors/factfilter/evaluator/InterpretExprAsFunction.scala
```diff
@@ -340,17 +354,19 @@ object InterpretExprAsFunction {
 
   final case class Input[F[_], V](
     value: F[V],
-    evidence: Evidence,
+    evidence: Evidence, // TODO: Is this needed? Can't I just add the input value as evidence when it is made of Facts?
```
jeffmay: This seems needed because it must be possible to select a value (like `0.8`) from a fact (like `ProbabilityToUse().scores("weightloss")`) without losing the fact it came from.

My hope was that I would be able to combine the evidence at a higher level in the expression once the value has been used, but I feel like there are too many edge cases to handle there and I'm not sure it is possible. However, with the current input.evidence, it is hard to know when to use it. I think there just needs to be a lot more unit tests around expected behavior before I'll know whether my hypothesis that we can handle this at some higher level expression can be proven or disproven.

jeffmay: I'm gonna remove this comment for now as I am doubting its possibility again and it isn't relevant to this PR.

core/src/main/scala/com/rallyhealth/vapors/factfilter/evaluator/InterpretExprAsFunction.scala
paul-a-kennedy-rally: yeah...

jeffmay:
Just as a note. I am planning to rework how evidence tracking is done. A lot of this work will have to be revised, but as an initial proof of concept implementation, I believe this is sufficient.

darmbrus (approved):

core/src/main/scala/com/rallyhealth/vapors/factfilter/data/FactOrFactSet.scala
paul-a-kennedy-rally: I agree that there really needs to be better names in this.  I know that your development is a bit of a hydra-model right now, so I don't know how painful this will be to do in the middle.

But this will need to be done before too much longer.

core/src/main/scala/com/rallyhealth/vapors/factfilter/evaluator/InterpretExprAsFunction.scala
paul-a-kennedy-rally: Yes, please do!

paul-a-kennedy-rally (approved):
There's a renaming armgeddon coming, but no immediate blockers.

core/src/main/scala/com/rallyhealth/vapors/factfilter/data/FactOrFactSet.scala
jeffmay: I think there may be some misunderstanding about this type. It should not be referenced directly. It only exists as a "magnet type", which is basically a target for implicit conversion that you can put on a method (can also be as varargs) that allows you to define various implicit conversions for different known types. This allows you to avoid method overloading, as well as allowing the implicit resolution system to prove type correctness before dropping the unnecessary type parameters (depending on what you keep in the magnet type). In my case, I'm just using this magnet type to allow calling the `Evidence.apply()` method with any variable combination of single facts, list of facts, set of facts, etc. without needing to define separate apply methods. In addition, this same magnet type is used in the `FactTable.apply` method to allow the same types of arguments to be passed.

core/src/main/scala/com/rallyhealth/vapors/factfilter/evaluator/InterpretExprAsFunction.scala
jeffmay: I think I'm gonna wait until after I rework the evidence tracking to document it, since it is fairly broken right now.

PR #28, opened 2020-12-18
Add OutputWithinSet expression and unit tests

jeffmay:
- Add OutputWithinSet operator for checking whether a given value exists within a given `Set` of values
- Add unit tests to verify

### DIFF
--- a/core/src/main/scala/com/rallyhealth/vapors/factfilter/dsl/ExprBuilder.scala
+++ b/core/src/main/scala/com/rallyhealth/vapors/factfilter/dsl/ExprBuilder.scala
@@ -4,6 +4,7 @@ import cats.{FlatMap, Foldable, Functor, Id, Order}
 import com.rallyhealth.vapors.core.algebra.Expr
 import com.rallyhealth.vapors.core.data.{NamedLens, Window}
 import com.rallyhealth.vapors.core.math.{Addition, Negative, Subtraction}
+import com.rallyhealth.vapors.factfilter.data.TypedFact
 
 import scala.collection.Factory
 
@@ -169,6 +170,13 @@ final class ValExprBuilder[V, R, P](returnOutput: Expr[Id, V, R, P])
     captureResult: CaptureP[Id, V, N[X], P],
   ): FoldOutExprBuilder[V, N, X, P] =
     new FoldOutExprBuilder(buildGetExpr(buildLens))
+
+  def value[X](
+    implicit
+    ev: R <:< TypedFact[X],
+    captureResult: CaptureResult[X],
+  ): ValExprBuilder[V, X, P] =
+    new ValExprBuilder(buildGetExpr[Id, X](_.field("value", _.value)))
 
   def in(accepted: Set[R])(implicit captureOutput: CaptureResult[Boolean]): ValExprBuilder[V, Boolean, P] =
     new ValExprBuilder(Expr.OutputWithinSet(returnOutput, accepted, captureOutput))
--- a/core/src/main/scala/com/rallyhealth/vapors/factfilter/dsl/ExprBuilder.scala
+++ b/core/src/main/scala/com/rallyhealth/vapors/factfilter/dsl/ExprBuilder.scala
@@ -170,6 +170,9 @@ final class ValExprBuilder[V, R, P](returnOutput: Expr[Id, V, R, P])
   ): FoldOutExprBuilder[V, N, X, P] =
     new FoldOutExprBuilder(buildGetExpr(buildLens))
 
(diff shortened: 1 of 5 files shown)


core/src/main/scala/com/rallyhealth/vapors/factfilter/dsl/ExprBuilder.scala
```diff
@@ -172,6 +173,16 @@ final class ValExprBuilder[V, R, P](returnOutput: Expr[Id, V, R, P])
   ): FoldOutExprBuilder[V, N, X, P] =
     new FoldOutExprBuilder(buildGetExpr(buildLens))
 
+  def value[X](
+    implicit
+    ev: R <:< TypedFact[X],
+    captureResult: CaptureResult[X],
+  ): ValExprBuilder[V, X, P] =
+    new ValExprBuilder(buildGetExpr[Id, X](_.field("value", _.value)))
```
kionka-rally, line 179: Why is the string `"value"` used here?

core/src/test/scala/com/rallyhealth/vapors/factfilter/evaluator/ConditionExprSpec.scala
```diff
@@ -0,0 +1,36 @@
...
+
+    "not find an asthma tag in a set that does not contain it" in {
+      val q = withFactsOfType(FactTypes.Tag).where {
+        _.exists {
+          _.value.in(Set("diabetes"))
+        }
+      }
+      val result = eval(JoeSchmoe.factTable)(q)
+      assert(!result.output.value)
+      // TODO: Should this contain the tags that did not match?
```
kionka-rally: I'm guessing not. I'm trying to think of a use case for that.

core/src/main/scala/com/rallyhealth/vapors/factfilter/dsl/ExprBuilder.scala
jeffmay: This is because all `Fact`s have a `.value` field that contains the value of the `Fact`?

core/src/test/scala/com/rallyhealth/vapors/factfilter/evaluator/ConditionExprSpec.scala
jeffmay: The only problem is... an answer without evidence is equivalent to no answer. So if I don't have evidence for the existence of a tag, I don't have evidence of the absence of a tag... That is just how it is.

I'm just not sure how this will work with `not(in(Set()))`. If I have some tags and none of the tag facts are in the set, shouldn't the evidence that the tags are "not in the Set" be the tags that are not in the Set?

jeffmay: Maybe I can make a special `not in` operator by using a boolean flag on the `SetContains` operator? Then when the Expr is negated, I can negate the contained node.

paul-a-kennedy-rally: I think that this is closer--there needs to be a way to represent the "not in " test (and supporting evidence)>

paul-a-kennedy-rally (commented):
Looks good so far, but I think that you should add more test cases to prove out the semantics that we need to support before it goes in or is put in production.

core/src/test/scala/com/rallyhealth/vapors/factfilter/evaluator/ConditionExprSpec.scala
jeffmay: I'm gonna create a separate ticket to return to this issue.

core/src/main/scala/com/rallyhealth/vapors/factfilter/dsl/ExprBuilder.scala
kionka-rally: I guess I'm still not clear why you are using a string literal in the expression builder.

kionka-rally (approved):

paul-a-kennedy-rally (approved):
LGTM

core/src/main/scala/com/rallyhealth/vapors/factfilter/dsl/ExprBuilder.scala
paul-a-kennedy-rally: Yeah, I agree that you have a magic constant here that's not really connect to the `Fact` (when it really should....)

darmbrus (approved):

PR #29, opened 2021-01-05
PRO-826 - Add support for wrapping HList with case class

jeffmay:
Add support for mapping multiple expression nodes into a case class.

Depends on:
- [x] https://github.com/jeffmay/vapors/pull/25


core/src/main/scala/com/rallyhealth/vapors/factfilter/evaluator/EvalExprAsOutputAndParam.scala
```diff
@@ -0,0 +1,105 @@
...
+import com.rallyhealth.vapors.core.algebra.Expr
+import com.rallyhealth.vapors.core.logic.{Conjunction, Disjunction, Negation}
+import com.rallyhealth.vapors.core.math.{Addition, Negative, Subtraction}
+import com.rallyhealth.vapors.factfilter.data.{ExtractBoolean, FactSet}
+import com.rallyhealth.vapors.factfilter.evaluator.InterpretExprAsFunction.{Input, Output}
+import shapeless.HList
+
+// TODO: Make this a proxy to a function?
+class EvalExprAsOutputAndParam[F[_] : Foldable, V, P]
+  extends Expr.Visitor[F, V, P, Lambda[r => Input[F, V] => (Output[r], List[Eval[P]])]] {
```
jeffmay: It sucks to have to write this class. I wonder if it is possible to write a simpler visitor that proxies to a function (or set of functions).

core/src/main/scala/com/rallyhealth/vapors/factfilter/evaluator/InterpretExprAsFunction.scala
```diff
@@ -231,6 +232,33 @@ final class InterpretExprAsFunction[F[_] : Foldable, V, P]
...
+    override def product[A, B](
+      fa: Fn[A],
+      fb: Fn[B],
+    ): Fn[(A, B)] = { input =>
+      val (a, aParams) = fa(input)
+      val (b, bParams) = fb(input)
+      // TODO: Product type should remove evidence if any arg has no evidence
+      (Output((a.value, b.value), a.evidence ++ b.evidence), aParams ::: bParams)
+    }
+  }
```
jeffmay: BLKR: I need to add unit tests for how to handle evidence.

core/src/main/scala/com/rallyhealth/vapors/factfilter/evaluator/InterpretExprAsFunction.scala
```diff
@@ -7,12 +7,13 @@ import com.rallyhealth.vapors.core.logic._
...
 import com.rallyhealth.vapors.factfilter.data._
 import com.rallyhealth.vapors.factfilter.evaluator.InterpretExprAsFunction.{Input, Output}
+import shapeless.HList
 
 import scala.collection.immutable.BitSet
 
 // TODO: Make R the last parameter?
 final class InterpretExprAsFunction[F[_] : Foldable, V, P]
-  extends Expr.Visitor[F, V, P, Lambda[r => Input[F, V] => ExprResult[F, V, r, P]]] {
+  extends Expr.Visitor[F, V, P, Lambda[r => Input[F, V] => ExprResult[F, V, r, P]]] { interpreter =>
```
jeffmay: This is unused.

core/src/main/scala/com/rallyhealth/vapors/core/algebra/HExprList.scala
```diff
@@ -0,0 +1,29 @@
+package com.rallyhealth.vapors.core.algebra
+
+import cats.syntax.all._
+import cats.{Functor, Semigroupal}
+import shapeless.{::, HList, HNil}
+
+// TODO: How to convert this back into an HList of Expr nodes?
+sealed trait NonEmptyExprHList[F[_], V, T <: HList, P] {
```
jeffmay: This is processed recursively using `EvalExprAsOutputAndParam` which calls back into `InterpretExprAsFunction` but strips off the `ExprResult` wrapper and returns the tuple of `(Output[R], P)`. This function is simple and generic enough that there is a `Functor` and `Semigroupal` definition for `EvalExprAsOutputAndParam`, so each of the results can be visited and mapped over to combine all the results back into an `HList` in the `NonEmptyExprHList.visit` method. This `HList` is then passed back to the original call from `InterpretExprAsFunction` and passed into the function that maps `HList` into the expected return type, then finally wrapped with `ExprResult.WrapOutput` and returned.

I'm not sure of a way to convert a `NonEmptyExprHList` into an `HList` of `Expr` nodes. That would require some type magic that I don't understand. We don't really have the `HList` result until we evaluate the expression, so it only seems possible to work backwards from there (which is what I do). I believe it is possible to do some type magic to make this possible, but I am not smart enough to figure it out. I'll ask on the forums.

core/src/main/scala/com/rallyhealth/vapors/factfilter/evaluator/EvalExprAsOutputAndParam.scala
jeffmay: BLKR: This definitely needs documentation.

core/src/main/scala/com/rallyhealth/vapors/factfilter/evaluator/InterpretExprAsFunction.scala
jeffmay: I'm punting on evidence tracking until later, so I think we can ignore this one for now.

core/src/test/scala/com/rallyhealth/vapors/factfilter/evaluator/WrapOutputSpec.scala
```diff
@@ -0,0 +1,36 @@
...
+      }
+      val result = eval(FactTable.empty)(q)
+      assertResult(Point2D(1, 2))(result.output.value)
+    }
+
+    "NOT allow wrapping three integers as a Point2D" in {
+      assertDoesNotCompile {
+        "wrap(const(1), const(2), const(3)).as[Point2D]"
+      }
+    }
```
jeffmay: BLKR: This needs more unit tests for different types of expressions and expressions that use the input type of `F[V]`.

core/src/main/scala/com/rallyhealth/vapors/factfilter/dsl/ExprBuilder.scala
```diff
@@ -54,6 +55,7 @@ final class AdditionBuilderOps[R : Addition](number: R) {
   def +[V, P](
     builder: ValExprBuilder[V, R, P],
   )(implicit
+    captureEmbed: builder.CaptureEmbed[R],
```
jeffmay: BLKR: Why is this needed?

core/src/main/scala/com/rallyhealth/vapors/factfilter/evaluator/InterpretExprAsSimpleOutputFn.scala
```diff
@@ -0,0 +1,43 @@
...
+        case (o, params) => (o.copy(value = f(o.value)), params)
+      }
+
+      override def product[A, B](
+        fa: SimpleOutputFn[A],
+        fb: SimpleOutputFn[B],
+      ): SimpleOutputFn[(A, B)] = { input =>
+        val (a, aParams) = fa(input)
+        val (b, bParams) = fb(input)
+        // TODO: Product type should remove evidence if any arg has no evidence
```
jeffmay: I'm ignoring this for now since we have to revisit evidence tracking anyway, but it might be worth encoding this now as a reminder. Do y'all think this makes sense? If there is no evidence for a single value of product type (think tuple or case class) then there is no evidence for the full product type instance?

paul-a-kennedy-rally (approved):
LGTM

core/src/main/scala/com/rallyhealth/vapors/core/algebra/HExprList.scala
jeffmay: I removed this TODO. I don't think it is possible at this level.

paul-a-kennedy-rally (approved):

core/src/main/scala/com/rallyhealth/vapors/factfilter/dsl/WrapExprSyntax.scala
```diff
@@ -0,0 +1,156 @@
...
+  extends AnyVal {
+
+  def as[R](
+    implicit
+    gen: Generic.Aux[R, L],
+    captureResult: CaptureP[F, V, R, P],
+  ): Expr[F, V, R, P] =
+    Expr.WrapOutput(exprHList, gen, captureResult)
+
+  def asHList(
```
jeffmay, line 142: Why not also add `.asTuple`?

darmbrus (approved):

PR #32, opened 2021-01-15
Add equals / not equals operators as well as set filtering operators and unit tests

jeffmay:
- Add `FilterOutput` expression node for filtering a collection of values down to a given set
- Add `OutputIsEmpty` expression node for converting a collection to a boolean, returning whether it is empty or not
- Update `OutputWithinWindow` to support `equalTo`, `===`, and `!==` operations
- Add `Window.equalTo` method that creates an inclusive range
- Add unit tests that ignore evidence tracking


paul-a-kennedy-rally (approved):
LGTM

darmbrus (approved):

PR #33, opened 2021-01-20
Revert to Seq instead of SortedSet for facts

jeffmay:
**Problem:**
When adding facts to the `FactTable`, the library was using the given `Order` for that type to determine duplicates. This was not the intended purpose for the defined `Order`, so it was causing some non-duplicate facts to get removed from the `FactTable`. Prior to this, we switched from using a `SortedSet` to using a `Set` to store the facts (for other reasons). Because of this, it would be easy to just switch to using a `Set` for storage, and rely on equality checking *only*. Now, when receiving the facts as an input, the sequence will be sorted by the `FactType.order`, but it will not remove any instances when the comparator returns 0. In addition a `.toSet` operation was added to the builder for backwards compatibility and possibly other benefits of using a `Set` in expressions.

**Fixes:**
- Use `Seq[TypedFact[T]]` input for `withFactsOfType` builder
- Add method to get a `SortedSet` of facts (returned as `Set`) from the builder

### DIFF
--- a/core/src/main/scala/com/rallyhealth/vapors/core/algebra/Expr.scala
+++ b/core/src/main/scala/com/rallyhealth/vapors/core/algebra/Expr.scala
@@ -129,7 +129,7 @@ object Expr {
     */
   final case class WithFactsOfType[T, R, P](
     factTypeSet: FactTypeSet[T],
-    subExpr: Expr[Set, TypedFact[T], R, P],
+    subExpr: Expr[Seq, TypedFact[T], R, P],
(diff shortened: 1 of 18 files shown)


kionka-rally (approved):

kionka-rally:
It kinda sounds like the `Order` instances were the ones that were wrong.

vinit-garg (approved):

paul-a-kennedy-rally (approved):
LGTM

PR #34, opened 2021-01-20
Combine the lenses of consecutive calls to .get() in ValExprBuilder

jeffmay:
When looking at the debug output of consecutive `.get` calls.

For example:
```scala
_.get(_.select(_.value)).getFoldable(_.scores)
```

It would be nice to see a single lens path: `_.value.scores`

This PR does so by matching on the previous node to see if it is a Select node, and then chains the lenses together and returns a single combined node. This optimizes the query while it is being built.

This PR also:

- Adds `Map` instances for foldable / filter to the ExprDsl
- Uses a better `.toString` for the standard `captureUnit` and `captureParamAndPass` instances (previously showed lambda + memory location, which was always the same and not helpful)

Depends on:
- [x] https://github.com/jeffmay/vapors/pull/33


paul-a-kennedy-rally (approved):
Nice!

PR #37, opened 2021-01-25
Convert FilterOutput to take a condition expression

jeffmay:
- Convert FilterOutput to take a condition expression
- Update unit tests to reflect the expanded scope of the filter operator
- Update the ExprBuilder methods

### DIFF
--- a/core/src/main/scala/com/rallyhealth/vapors/core/algebra/Expr.scala
+++ b/core/src/main/scala/com/rallyhealth/vapors/core/algebra/Expr.scala
@@ -257,8 +257,7 @@ object Expr {
 
   final case class FilterOutput[F[_], V, M[_] : Foldable : FunctorFilter, R, P](
     inputExpr: Expr[F, V, M[R], P],
-    validValues: Set[R],
-    valuesAsEvidence: Boolean,
+    condExpr: Expr[Id, R, Boolean, P],
     capture: CaptureP[F, V, M[R], P],
   ) extends Expr[F, V, M[R], P] {
     override def visit[G[_]](v: Visitor[F, V, P, G]): G[M[R]] = v.visitFilterOutput(this)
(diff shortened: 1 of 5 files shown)


core/src/test/scala/com/rallyhealth/vapors/factfilter/evaluator/FilterOutputSpec.scala
```diff
@@ -12,141 +12,136 @@ class FilterOutputSpec extends AnyWordSpec {
 
   "Expr.FilterOutput" when {
 
-    "using containsAny op" should {
+    "using with an 'OutputWithinSet' op" when {
```
jeffmay, line 13: It might be easier to review this file by just looking at the whole side-by-side. I rearranged the tests so that everything lives under this `"using with an 'OutputWithinSet' op"` and moved the `containsAny` syntactic sugar to the bottom of this section. This would leave me space to add tests for passing different condition expression operator operator -- namely, the `OutputWithinRange` operator.

paul-a-kennedy-rally (approved):
LGTM

core/src/main/scala/com/rallyhealth/vapors/core/algebra/Expr.scala
```diff
@@ -257,8 +257,7 @@ object Expr {
 
   final case class FilterOutput[F[_], V, M[_] : Foldable : FunctorFilter, R, P](
     inputExpr: Expr[F, V, M[R], P],
-    validValues: Set[R],
-    valuesAsEvidence: Boolean,
+    condExpr: Expr[Id, R, Boolean, P],
```
jeffmay, line 260: BLKR: I need to update the `ExprResult` to reflect this new sub-expression.

core/src/main/scala/com/rallyhealth/vapors/factfilter/evaluator/InterpretExprAsFunction.scala
```diff
@@ -178,6 +190,7 @@ final class InterpretExprAsFunction[F[_] : Foldable, V, P]
       mapFn(mapInput)
     }
     val allValues = allResults.map(_.output.value)
+    // TODO: Shouldn't I only look at the evidence of results that aren't empty?
```
jeffmay: Wait... this is the `.map` operator... technically, every input should have exactly 1 output, so there is no higher-kinded definition of an "empty" output. I should probably remove this TODO.

darmbrus (approved):

PR #39, opened 2021-02-04
Move all code to core package, leave aliases for source compatibility

jeffmay:
This PR cleans up a lot of TODOs regarding packaging. Everything in the vapors-core project now lives in the `com.rallyhealth.vapors.core` package. Some aliases were left in the old `com.rallyhealth.vapors.factfilter` package to make the transition easier.

I broke this PR out into a lot of bite-sized commits, so hopefully that helps.


core/src/main/scala/com/rallyhealth/vapors/factfilter/evaluator/package.scala
```diff
@@ -0,0 +1,30 @@
+package com.rallyhealth.vapors.factfilter
+
+import com.rallyhealth.vapors.core.interpreter
+
+package object evaluator {
+
+//  @deprecated("Use com.rallyhealth.vapors.core.interpreter.ExprInput instead.", "0.8.0")
```
paul-a-kennedy-rally: I assume that this will be removed?  (I'm going commit by commit...)

paul-a-kennedy-rally (approved):
LGTM

Having the discrete changes nicely articulated made this a joy to review.  Thank you for doing this.

core/src/main/scala/com/rallyhealth/vapors/factfilter/evaluator/package.scala
jeffmay: My process for each commit was:
1. I added the aliases to make sure the code still compiled (mostly)
2. I removed the aliases and found all the spots that failed to compile
3. Updated the code to use the new package paths
4. Uncommented the aliases and committed the change

... however, it looks like I forgot the last step on this commit.

darmbrus (approved):

PR #40, opened 2021-02-11
Added Expr.ZipOutput operator

jeffmay:
- Updated NonEmptyExprHList to add a higher-kinded param
- Renamed visit method to visitProduct and use Semigroupal / Functor
- Added visitZippedToShortest and use Align / FunctorFilter
- Updated detailed documentation of NonEmptyExprHList methods
- Updated wrap methods to use Id type constructor
- Added ZipOutputSpec for testing various cases
- Convert WrapOutputSpec to AnyFreeSpec and organize similar to ZipOutputSpec
- Updated unit tests to share simple tag examples


core/src/main/scala/com/rallyhealth/vapors/core/dsl/ZipExprSyntax.scala
```diff
@@ -0,0 +1,81 @@
...
+
+  def zip[F[_], V, M[_] : Align : Functor, E1, E2, E3, E4, P](
+    e1: Expr[F, V, M[E1], P],
+    e2: Expr[F, V, M[E2], P],
+    e3: Expr[F, V, M[E3], P],
+    e4: Expr[F, V, M[E4], P],
+  ): ZipExprHListWrapper[F, V, M, E1 :: E2 :: E3 :: E4 :: HNil, P] =
+    new ZipExprHListWrapper(e1 :: e2 :: e3 :: NonEmptyExprHList.tailK(e4))
+
+  def zip[F[_], V, M[_] : Functor : Semigroupal, E1, E2, E3, E4, E5, P](
```
jeffmay: BLKR: Remove the type-classes on this. They are unused.

core/src/main/scala/com/rallyhealth/vapors/core/dsl/ZipExprSyntax.scala
```diff
@@ -0,0 +1,81 @@
...
+    e1: Expr[F, V, M[E1], P],
+    e2: Expr[F, V, M[E2], P],
+    e3: Expr[F, V, M[E3], P],
+    e4: Expr[F, V, M[E4], P],
+    e5: Expr[F, V, M[E5], P],
+  ): ZipExprHListWrapper[F, V, M, E1 :: E2 :: E3 :: E4 :: E5 :: HNil, P] =
+    new ZipExprHListWrapper(e1 :: e2 :: e3 :: e4 :: NonEmptyExprHList.tailK(e5))
+}
+
+final class ZipExprHListWrapper[F[_], V, M[_], L <: HList, P](private val exprHList: NonEmptyExprHList[F, V, M, L, P])
```
jeffmay: I wonder if this should be more generically named than "zip". I could apply any higher-kinded operation to all the expressions.

Operations could be concatenation or permutations for example. In the case of concatenation, wouldn't the elements need to be of the same type? If so, this would not be a good interface for that outcome? In the case of permutations, this would be a good interface, but when would we ever need that operator?

If I try to come up with more generic names, I can think of `wrapEach` ("each" isn't really a common naming suffix for operations on collections and it kind of implies zipping), `wrapEvery` (same comment as wrapEach, but implies every element of the output is visited, which is probably worse), `wrapFoldable` (aligns with other spots like `.getFoldable`, except the types are not required to be `Foldable`), `group` (a new term, would not show up next to `wrap` in autocompletion, but short and fairly clear). I'm not happy with any of these more generic names. The would also require the method names in this class change to `.zip` for clarity.

jeffmay: I suppose I could change `wrap` to `group` (so it would be `group(const(1), const("A")).asTuple` and call this method `groupEach`:

```
groupEach(const(List(1, 2, 3)), const("A", "B")).zippedToShortest.asTuples
// produces List((1, "A"), (2, "B"))
groupEach(const(List(1, 2, 3)), const("A", "B")).permutations.asTuples
// produces List((1, "A"), (1, "B"), (2, "A"), (2, "B"))
```

core/src/main/scala/com/rallyhealth/vapors/core/algebra/HExprList.scala
```diff
@@ -1,28 +1,133 @@
...
+    * @example NonEmptyExprHList.of(const("A"), const(1)).visitProduct(g) == const("A" :: 1 :: HNil).visit(g)
+    *
+    * @note If given a [[List]], this will combine the resulting lists of each expression into a list of
+    *       every permutation of HList that can be produced from the output lists.
+    *
+    * @example NonEmptyExprHList.of(const(List("A", "B")), const(List(1, 2))).visitProduct(g) == const(
+    *            List("A" :: 1 :: HNil, "A" :: 2 :: HNil, "B" :: 1 :: HNil, "C" :: 2 :: HNil)
+    *          ).visit(g)
+    *
+    * @example NonEmptyExprHList.of(const(List("A")), const(Nil)).visitProduct(g) == const(Nil).visit(g)
```
paul-a-kennedy-rally, line 40: I like that you have examples, but maybe these simple examples could also be put in tests (with a link in the docs here)?

You'd get:
- stable link from this documentation to test
- guarantee of test code revealing supported status.
- an indication should the text examples the executable documentation drift.

core/src/main/scala/com/rallyhealth/vapors/core/algebra/HExprList.scala
```diff
@@ -1,28 +1,133 @@
...
+    * Visits each expression in the HList, applies the given visitor, zips over every element in the resulting
+    * functor-wrapped values (stopping at the length of the first sequence to run out of elements), and then
+    * combines the elements into an HList.
+    *
+    * @example NonEmptyExprHList.of(const(List("A", "B", "C"), const(List(1, 2, 3, 4, 5))).visitZippedToShortest(g) ==
+    *            const(List("A" :: 1 :: HNil, "B" :: 2 :: HNil, "C" :: 3 :: HNil)).visit(g)
+    *
+    * @example NonEmptyExprHList.of(const(List("A", "B", "C"), const(Nil)).visitZippedToShortest(g) ==
+    *            const(Nil).visit(g)
+    */
```
paul-a-kennedy-rally, line 59: Same thing here...  The R language has had support for "runnable examples" (and _maintained_ literate programming tooling) tools for as long as I can remember which makes verification that your documentation stays up to date pretty easy,

core/src/main/scala/com/rallyhealth/vapors/core/lens/NamedLens.scala
```diff
@@ -25,6 +28,36 @@ object NamedLens {
...
+//    def at(
+//      n: Nat,
+//    )(implicit
+//      at: At[L, n.N],
+//      toInt: ToInt[n.N],
+//    ): NamedLens[A, at.Out] =
+//      lens.copy(
+//        path = lens.path.atKey(Nat.toInt[n.N]),
+//        get = lens.get.andThen(at.apply(_)),
+//      )
```
jeffmay: Remove if unused.

core/src/main/scala/com/rallyhealth/vapors/core/dsl/WrapEachExprSyntax.scala
```diff
@@ -0,0 +1,164 @@
...
+  ): WrapEachExprHListWrapper[F, V, M, E1 :: E2 :: HNil, P] =
+    new WrapEachExprHListWrapper(e1 :: NonEmptyExprHList.tailK(e2))
+
+  def wrapEach[F[_], V, M[_], E1, E2, E3, P](
+    e1: Expr[F, V, M[E1], P],
+    e2: Expr[F, V, M[E2], P],
+    e3: Expr[F, V, M[E3], P],
+  ): WrapEachExprHListWrapper[F, V, M, E1 :: E2 :: E3 :: HNil, P] =
+    new WrapEachExprHListWrapper(e1 :: e2 :: NonEmptyExprHList.tailK(e3))
+
```
paul-a-kennedy-rally, line 22: Not a blocker, but a "I wonder"--is there a better way to generalize this with a macro?

paul-a-kennedy-rally (approved):

darmbrus (approved):

core/src/main/scala/com/rallyhealth/vapors/core/algebra/HExprList.scala
jeffmay: I can try adding https://github.com/ThoughtWorksInc/sbt-example so that sbt runs the tests in the example tags.

core/src/main/scala/com/rallyhealth/vapors/core/dsl/WrapEachExprSyntax.scala
jeffmay: Looks like I have [some options](http://yefremov.net/blog/scala-code-generation/) here. Maybe I can try something similar to [the way cats handles different arity](https://github.com/typelevel/cats/blob/master/project/Boilerplate.scala).

PR #41, opened 2021-02-15
Remove the F[_] parameter from Expr

jeffmay:
I had an epiphany about the `Expr`. If every `Expr` node takes an `inputExpr`, then there is really no benefit to constraining the input type. You can always pass the `Expr.ReturnInput` to move a higher-kinded parameter from the input side to the output side, where you can constrain the output with an implicit. This vastly simplifies the code without reducing expressiveness. It might also resolve some issues with implicit resolution when the type parameter was `Id[V]`

**Dependencies**

- [x] https://github.com/jeffmay/vapors/pull/40

### DIFF
--- a/core/src/main/scala/com/rallyhealth/vapors/core/dsl/ExprDsl.scala
+++ b/core/src/main/scala/com/rallyhealth/vapors/core/dsl/ExprDsl.scala
@@ -176,17 +176,25 @@ trait ExprDsl extends WrapExprSyntax with WrapEachExprSyntax {
   ): Expr.ExistsInOutput[V, M, U, P] =
     Expr.ExistsInOutput(inputExpr, condExpr, capture)
 
+  def returnInput[V, P](
+    implicit
+    capture: CaptureP[V, V, P],
+  ): Expr.ReturnInput[V, P] =
+    Expr.ReturnInput(capture)
+
+  @deprecated("Use returnInput instead", "0.10.0")
   def returnInputFoldable[V, P](
     implicit
     capture: CaptureP[V, V, P],
   ): Expr.ReturnInput[V, P] =
     Expr.ReturnInput(capture)
 
+  @deprecated("Use returnInput instead", "0.10.0")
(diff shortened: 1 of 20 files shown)


core/src/main/scala/com/rallyhealth/vapors/core/interpreter/ExprInput.scala
```diff
@@ -8,7 +8,7 @@ final case class ExprInput[V](
   factTable: FactTable,
 ) {
 
-  // TODO: Remove
+  @deprecated("Use withValue instead.", "0.10.0")
```
paul-a-kennedy-rally, line 11: Do you want to include a target removal version?  I think that we should start doing this.

core/src/main/scala/com/rallyhealth/vapors/core/dsl/ExprDsl.scala
```diff
@@ -176,17 +176,25 @@ trait ExprDsl extends WrapExprSyntax with WrapEachExprSyntax {
   ): Expr.ExistsInOutput[V, M, U, P] =
     Expr.ExistsInOutput(inputExpr, condExpr, capture)
 
+  def returnInput[V, P](
+    implicit
+    capture: CaptureP[V, V, P],
+  ): Expr.ReturnInput[V, P] =
+    Expr.ReturnInput(capture)
+
+  @deprecated("Use returnInput instead", "0.10.0")
```
paul-a-kennedy-rally, line 185: (same thing here?  Target removal version?)

paul-a-kennedy-rally (approved):

darmbrus (approved):

core/src/main/scala/com/rallyhealth/vapors/core/interpreter/ExprInput.scala
jeffmay: I don't really know. I might remove it in 1.0.0, or I might remove it before then.

PR #44, opened 2021-02-19
Add Expr.ConcatOutput and tests

jeffmay:
- Concatenates monoids form multiple expressions into a single expression of the monoid result


core/src/main/scala/com/rallyhealth/vapors/core/dsl/ExprDsl.scala
```diff
@@ -77,6 +77,9 @@ trait ExprDsl extends WrapExprSyntax with WrapEachExprSyntax {
   ): Expr.Embed[V, R, P] =
     Expr.Embed(expr, captureResult)
 
+  def concat[V, M[+_], R, P](expressions: Expr[V, M[R], P]*): ConcatOutputExprBuilder[V, M, R, P] =
```
jeffmay: BLKR: I'm not sure if this restriction on M[] to be covariant is necessary. If not, I should remove it.

darmbrus (approved):

paul-a-kennedy-rally (approved):

PR #48, opened 2021-03-27
PRO-1015 - Add FoldOutput operator and unit tests

jeffmay:
A quick fix for supporting the `fold` operator

### DIFF
--- a/core/src/test/scala/com/rallyhealth/vapors/core/interpreter/FoldOutputSpec.scala
+++ b/core/src/test/scala/com/rallyhealth/vapors/core/interpreter/FoldOutputSpec.scala
@@ -1,0 +1,29 @@
+package com.rallyhealth.vapors.core.interpreter
+
+import com.rallyhealth.vapors.core.data.FactTable
+import com.rallyhealth.vapors.core.dsl._
+import org.scalatest.freespec.AnyFreeSpec
+
+class FoldOutputSpec extends AnyFreeSpec {
+
+  "Expr.FoldOutput" - {
+
+    "fold a list of ints into its sum" in {
+      val query = const(List(1, 2, 3)).withOutputFoldable.fold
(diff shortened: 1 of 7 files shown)


darmbrus (approved):

core/src/test/scala/com/rallyhealth/vapors/core/interpreter/FoldOutputSpec.scala
```diff
@@ -0,0 +1,29 @@
...
+import org.scalatest.freespec.AnyFreeSpec
+
+class FoldOutputSpec extends AnyFreeSpec {
+
+  "Expr.FoldOutput" - {
+
+    "fold a list of ints into its sum" in {
+      val query = const(List(1, 2, 3)).withOutputFoldable.fold
+      val result = eval(FactTable.empty)(query)
+      assertResult(6)(result.output.value)
```
paul-a-kennedy-rally, line 14: Is it worthwhile to have a different Monoid supplied to change the behaviour?

paul-a-kennedy-rally (approved):
LGTM

core/src/test/scala/com/rallyhealth/vapors/core/interpreter/FoldOutputSpec.scala
jeffmay: I suppose this is just baked into cats now, so yea, you would have to supply your own monoid for `Int` if you wanted to do something other than addition. Maybe I could define a specific `.sum` and `.product` that are strict about the type of monoid they use?

jeffmay: Either way, this PR meets our needs for now, so I am going to move forward and we can revisit this before v1

PR #55, opened 2021-04-08
Use IDE package prefix feature and organize all imports

jeffmay:
This uses the InteliJ IDEA feature of [idePackagePrefix](https://blog.jetbrains.com/scala/2020/11/26/enhanced-package-prefixes/) to cut down on the depth of package directories and import paths.

~~I'm still considering whether having a `core` package is preferable... I might do something similar to `cats` and use a `kernel` package, but put all my imports into the `vapors` package. So anything I want to hide from the general import of `vapors._` can go into the kernel package, and everything else can sit one level down at the `vapors` package. In other words, use the `idePackagePrefix := Some("com.rallyhealth")` but put all the core things into the `vapors` package (except for things that should be hidden from the global import of `vapors._`. I would have to create another set of aliases to migrate existing code, but I would rather do this early to give time to downstream implementations to make the switch.~~

I decided to remove the `core` package from the imports as well. The `core` project will have no prefix (besides `com.rallyhealth.vapors`). If we break out any sort of "kernel" in the way that `cats` does, then we would create the aliases from `vapors.x` in the `core` project to point to the appropriate `vapors.kernel.x`. Since we don't have this, I am removing the `core` package and flattening it into `vapors`. Some things do not have aliases because they should not be used directly anyway.


darmbrus (approved):

build.sbt
```diff
@@ -31,9 +31,14 @@ ThisBuild / packageDoc / publishArtifact := false
...
-def commonProject(dir: String): Project = {
+def commonProject(
+  dir: String,
+  projectPrefix: String = "",
+): Project = {
+  val packagePrefix = s"com.rallyhealth${if (projectPrefix.isEmpty) "" else s".$projectPrefix"}"
   Project(dir, file(dir)).settings(
     name := s"vapors-$dir",
+    idePackagePrefix := Some(packagePrefix),
   )
```
paul-a-kennedy-rally, line 42: I wonder about symmetry on this (and maybe you've already tested this)?

Should the package prefix end with a dot or not?  When this translates to paths, are we sure that we're not going to end up with "/some/prefix//and/something/added" or worse "/some/prefixand/something/added"?

paul-a-kennedy-rally (approved):
I'm not sure I'm convinced about whether this is a good change or not?  I see how this makes it easier to re-home things if that's what you want.

I'm willing to see how this goes :)

PR #58, opened 2021-04-13
Create valuesOfType operator and use it in the test code

jeffmay:


paul-a-kennedy-rally (approved):

jeffmay:
Unfortunately this breaks some things with evidence tracking / justification. There is already a lot of know issues with the justification process, and there is a plan to fix it that goes well beyond the scope of this PR. Just be cautious if you are relying on evidence tracking / justification, because the behavior between `factsOfType(x).map(something(_.value))` can be different than `valuesOfType(x).map(something(_))`

darmbrus (approved):

PR #60, opened 2021-05-17
Update README.md to use the latest algebra / code examples.

jeffmay:


paul-a-kennedy-rally (approved):
I would really prefer if more of this were in the codebase and accessible from the IDE. 

Javadoc supported for aeons a way in which you could have overview docs embedded (and accessible in IDEs) as an integral part of the doc build (and codebase).

jeffmay:
> I would really prefer if more of this were in the codebase and accessible from the IDE.
> 
> Javadoc supported for aeons a way in which you could have overview docs embedded (and accessible in IDEs) as an integral part of the doc build (and codebase).

I updated some of the code documentation. Unfortunately ScalaDoc generation is broken because the way that relative code links are done in comments is a little bit strange and fails to compile in most cases. I'd be open to any PR that can fix that.

core/src/main/scala/vapors/dsl/ExprDsl.scala
```diff
@@ -17,6 +17,11 @@ object ExprDsl extends ExprDsl
 // TODO: Remove methods that are not useful anymore and move less-useful methods to a separate object
 trait ExprDsl extends TimeFunctions with WrapExprSyntax with WrapEachExprSyntax {
 
+  /**
+    * Evaluates the given [[RootExpr]] with the given [[FactTable]] to produce the [[ExprResult]].
+    *
+    * To access
```
darmbrus: `To access`...?

darmbrus (approved):
One comment about something that looks like an unfinished sentence. Other than that, LGTM.

paul-a-kennedy-rally:
> > I would really prefer if more of this were in the codebase and accessible from the IDE.
> > Javadoc supported for aeons a way in which you could have overview docs embedded (and accessible in IDEs) as an integral part of the doc build (and codebase).
> 
> I updated some of the code documentation. Unfortunately ScalaDoc generation is broken because the way that relative code links are done in comments is a little bit strange and fails to compile in most cases. I'd be open to any PR that can fix that.

I tried to build the scaladoc and it stops after 50 errors, so there are a lot of issues just to even get started now :(

paul-a-kennedy-rally (approved):

PR #61, opened 2021-06-28
Allow embedding an expression for target of comparison

jeffmay:
This is kind of a quick fix for allowing comparing the result of two expressions (as opposed to just a single expression against a constant). I think there may be a more reusable and elegant way to solve this, but I'm putting that off until I have more time to work on it.

**Previously**
```scala
const(5) < 7 // compiled
const(5) < const(7) // didn't compile
```

**After this change**
```scala
const(5) < 7 // compiles, but requires more captureP params
const(5) < const(7) // compiles now too
```


core/src/main/scala/vapors/algebra/Expr.scala
```diff
@@ -390,6 +392,14 @@ object Expr {
     override def visit[G[_]](v: Visitor[V, P, G]): G[R] = v.visitFoldOutput(this)
   }
 
+  final case class WrapOutput[V, L, R, P](
```
paul-a-kennedy-rally, line 403: Can you document this class (reason needed, purpose)?

core/src/main/scala/vapors/data/Window.scala
```diff
@@ -28,6 +29,35 @@ object Window {
...
+  implicit def genericIorBounds[A : Order]: Generic.Aux[Window[A], Ior[Bounded.Above[A], Bounded.Below[A]] :: HNil] =
+    new Generic[Window[A]] {
+      override type Repr = Ior[Bounded.Above[A], Bounded.Below[A]] :: HNil
+      override def to(windows: Window[A]): Ior[Bounded.Above[A], Bounded.Below[A]] :: HNil =
+        windows.bounds :: HNil
+      override def from(bounds: Ior[Bounded.Above[A], Bounded.Below[A]] :: HNil): Window[A] =
+        Window.fromBounds(bounds.head)
+    }
+
+//  implicit def genericIorValues[A : Order]: Generic.Aux[Window[A], Ior[A, A]] = new Generic[Window[A]] {
```
paul-a-kennedy-rally: Do you want to remove this?

paul-a-kennedy-rally (approved):
Some docs would be appreciated :)

darmbrus (approved):

PR #62, opened 2021-06-28
Avoid overloaded method for withinWindow and deprecate other non-Expr overloaded methods

jeffmay:


paul-a-kennedy-rally (approved):
I like the informative deprecation warnings.

darmbrus (approved):

PR #64, opened 2021-09-01
[NOJIRA] Add documentation to 'between' functions.

paul-a-kennedy-rally:


vinit-garg (approved):

core/src/main/scala/vapors/algebra/Expr.scala
```diff
@@ -28,7 +28,7 @@ import scala.collection.MapView
   * R = Return type
   * P = Captured param
   */
-sealed abstract class Expr[V, R, P] {
+sealed abstract class Expr[V, R, P] {Coursera
```
jeffmay: This looks like a bug.

jeffmay (changes requested):
Looks like you have a bug in the code. Please run tests to confirm everything is working. This is still a prototype branch, so CI is not enabled here.

core/src/main/scala/vapors/algebra/Expr.scala
paul-a-kennedy-rally: Yeah--was trying to clean spam out of my inbox.

paul-a-kennedy-rally:
> Looks like you have a bug in the code. Please run tests to confirm everything is working. This is still a prototype branch, so CI is not enabled here.

Tested and passes.

jeffmay (approved):

PR #65, opened 2021-09-02
Remove deprecated value operator syntax

jeffmay:

PR #66, opened 2021-09-02
Remove deprecated warnings from the arithmetic operators

jeffmay:
I attempted to roll out the removal of the deprecated methods from https://github.com/jeffmay/vapors/pull/65 in the Rally code base and type inference was not improved by removing these methods. I have plans to improve type inference with a better DSL and algebra in the future, but for now, we should remove these deprecation warnings.


darmbrus (approved):

paul-a-kennedy-rally (approved):
I appreciate this update.

Could we stamp a version and update kerna?

PR #67, opened 2021-11-15
Add Expr.ExponentiateOutputs and unit tests

jeffmay:


core/src/test/scala/vapors/interpreter/VaporsEvalTestHelpers.scala
```diff
@@ -97,8 +97,18 @@ object VaporsEvalTestHelpers {
         },
         expectedValue => {
           val result = eval(FactTable.empty)(query)
-          assertResult(expectedValue) {
-            result.output.value
+          expectedValue match {
+            case num: Double if num.isNaN =>
+              result.output.value match {
+                case obs: Double =>
+                  assert(obs.isNaN, "Expected output to be NaN")
```
jeffmay, line 104: `NaN` is pretty much the only value that isn't equal to itself, so I feel like a special case is warranted here. How this would play out in plain Scala code would be effectively the same as the result you would get from vapors, so handling `NaN` in the resulting output is still the responsibility of the library user.

darmbrus (approved):

paul-a-kennedy-rally (approved):
LGTM

PR #68, opened 2021-11-16
Add collection operations and support for justified result DSL (part 2 of 3+)

jeffmay:
* Make use of `Justified` wrapper class in the DSL traits
  * Allow chaining justification through a `Justified.ByInference` case class
* Add a `CustomFunction` operation for doing simple background conversions that would be difficult to implement in pure expressions (like traversing the values of a justified sequence)
* Fix issues attaching debug function
  * Use type-level programming to infer accurate type information
  * Actually invoke the attached debugger with the correct types in both the standard engine and simple engine
* Add comparison operations (`>`, `<=`, etc using `WithinWindow` expression)
* Add `.get` and `_.select` operations using a `VariantLens` (too be renamed after replacing the old invariant `NamedLens`)
* Add `.exists` and `.forall` expression builder functions to all DSLs (justified / unwrapped & simple / standard)


circe-v1/src/main/scala/dsl/circe/HasEncoder.scala
```diff
@@ -0,0 +1,20 @@
+package com.rallyhealth.vapors.v1
+
+package dsl.circe
+
+import io.circe.Encoder
+
+// TODO: Figure out how to mix and match DSLs
```
jeffmay: I think I can remove this TODO. This PR already utilizes this trait.

circe-v1/src/main/scala/dsl/circe/UnwrappedCirceBuildExprDsl.scala
```diff
@@ -0,0 +1,11 @@
+package com.rallyhealth.vapors.v1
+
+package dsl.circe
+
+import dsl.UnwrappedBuildExprDsl
+
+// TODO: Do JustifedExprDsl
```
jeffmay: Do this TODO

core-v1/src/main/scala/data/Fact.scala
```diff
@@ -105,11 +105,12 @@ object TypedFact {
...
-
-  def value[A]: NamedLens[TypedFact[A], A] = {
-    NamedLens[TypedFact[A], A](DataPath.empty.atField("value"), _.value)
-  }
+  // TODO: Switch to VariantLens
+//  def lens[A]: NamedLens.Id[TypedFact[A]] = NamedLens.id[TypedFact[A]]
+//
+//  def value[A]: NamedLens[TypedFact[A], A] = {
+//    NamedLens[TypedFact[A], A](DataPath.empty.atField("value"), _.value)
+//  }
```
jeffmay: I don't think I really use these methods. I mostly rely on the expression builders in both algebras / DSLs.

core-v1/src/main/scala/data/FactType.scala
```diff
@@ -106,9 +106,10 @@ object FactType {
     }
   }
 
-  implicit class ToLens[T](@nowarn private val ft: FactType[T]) extends AnyVal {
-    def lens: NamedLens[TypedFact[T], T] = TypedFact.value[T]
-  }
+  // TODO: Convert to NamedLens
+//  implicit class ToLens[T](@nowarn private val ft: FactType[T]) extends AnyVal {
+//    def lens: NamedLens[TypedFact[T], T] = TypedFact.value[T]
+//  }
```
jeffmay: Remove if unused.

core-v1/src/main/scala/algebra/ExprConverter.scala
```diff
@@ -0,0 +1,50 @@
...
+object ExprConverter {
+
+  // TODO: Capture type information about R for debugging?
+  private final class Impl[-I, +O](
+    override val conversionType: String,
+    convert: I => O,
+  ) extends ExprConverter[I, O] {
+    override def apply(in: I): O = convert(in)
+    override val toString: String = s"ExprConverter.$conversionType"
+  }
```
jeffmay: The type name is now `O` and I am not sure I will actually use this converter because I have to handle wrapped types which becomes a lot more complicated. This pattern only handles types of kind `*` and not `* -> *` or constants.

core-v1/src/main/scala/algebra/Extract.scala
```diff
@@ -0,0 +1,26 @@
...
+
+import data.Justified
+
+import cats.Id
+
+// TODO: Replace with cats Monoid or ZIO Prelude?
+trait Extract[F[_]] {
+
+  def extract[A](fa: F[A]): A
+}
```
jeffmay: There is a typeclass for `Extract` in alleycats... not sure if I want to pull in all of alleycats though...

core-v1/src/main/scala/algebra/WindowComparable.scala
```diff
@@ -0,0 +1,60 @@
...
+
+  // TODO: Should this take OP[V] and OP[Window[V]]? What about the F[_] wrapper?
+  //       For now, I am only taking the OP[V] because if the goal is to serialize the value and the window,
+  //       then that is the missing piece.
+  def withinWindow[V](
+    value: F[V],
+    window: F[Window[V]],
+  )(implicit
+    opV: OP[V],
+  ): F[Boolean]
```
jeffmay: I'm not really using the `OP[V]` here, but the idea is that if the `OP` provides an encoder for `V`, then I can serialize the window and the value as json (or whatever type of encoder it is). I'm not really using it right now though. I'm gonna leave it until it becomes an impediment.

core-v1/src/main/scala/data/Justified.scala
```diff
@@ -2,34 +2,126 @@ package com.rallyhealth.vapors.v1
...
+
+  // TODO: Allow a lens for map operations?
+  final case class ByFact[V](fact: TypedFact[V]) extends Justified[V] {
+    override val evidence: SomeEvidence = SomeEvidence(NonEmptySet.of(fact))
+    override def value: V = fact.value
+    override val reason: String = s"[fact: ${fact.typeInfo.name}=${fact.value}]"
+    override def visit[G[+_]](v: Visitor[G]): G[V] = v.visitFact(this)
+    override def configs: Seq[(String, Option[String])] = Seq.empty
+    override def productPrefix: String = "Justified.ByFact"
+  }
```
jeffmay: I achieve the same effect of a map operation using the `ByInference` class... so I'm not sure if this TODO is needed. Best argument I can make in favor of adding the lens here is that it might reduce the depth of the justification tree because all `valuesOfType` operations return a sequence of justified values by facts, and most operations on these individual facts will start with a `.get` selector (unless the fact is already the correct type of value, in which case the lens will be the identity lens). I'm not sure it is worth it.

I should either leave a more detailed comment for future experimentation or just remove the comment and add it back if needed.

core-v1/src/main/scala/data/Justified.scala
```diff
@@ -2,34 +2,126 @@ package com.rallyhealth.vapors.v1
...
+  final case class ByInference[V](
+    reason: String,
+    value: V,
+    sources: NonEmptyList[Justified[Any]], // TODO: Should this be a NonEmptySet? How to sort it?
   ) extends Justified[V] {
-    override def withValue[U](value: U): Justified[U] = copy(value = value)
-    override def evidenceOrNone: Evidence = evidence
+    override def visit[G[+_]](v: Visitor[G]): G[V] = v.visitInference(this)
+    override lazy val configs: Seq[(String, Option[String])] = sources.map(_.configs).reduce // TODO: should this be sorted?
+    override lazy val evidence: Evidence = sources.foldLeft(Evidence.none)(_ | _.evidence) // TODO: Is this even valid?
```
jeffmay: These seem like important TODOs. We aren't using justification yet, and I don't really have time to experiment with these, so I'm gonna leave them for now.

core-v1/src/main/scala/algebra/CompareWrapped.scala
```diff
@@ -0,0 +1,76 @@
...
+
+  def map[A, B](fa: F[A])(ab: A => B): F[B]
+
+  def wrapConst[A](value: A): F[A]
+
+  def compare[V](
+    value: F[V],
+    window: F[Window[V]],
+  ): F[Boolean]
+}
```
jeffmay: `Extract` and `WrapConst` already handle 2 of the methods in here. I should do these TODOs before merging to avoid confusion.

core-v1/src/main/scala/algebra/CompareWrapped.scala
```diff
@@ -0,0 +1,76 @@
...
+
+object CompareWrapped {
+
+  @inline final def apply[F[+_] : CompareWrapped]: CompareWrapped[F] = implicitly
+
+  // Can't use cats.Id because this needs to be covariant
+  type NoWrapper[+A] = A
```
jeffmay: I can use `shapeless.Id` however, because it is covariant.

core-v1/src/main/scala/algebra/CompareWrapped.scala
```diff
@@ -0,0 +1,76 @@
...
+      override def wrapConst[V](value: V): Justified[V] = Justified.byConst(value)
+
+      override def compare[V](
+        value: Justified[V],
+        window: Justified[Window[V]],
+      ): Justified[Boolean] = {
+        val comparison = Window.showWindowWithTerm[V]("_")(Show.fromToString).show(window.value)
+        val isWithinWindow = window.value.contains(value.value)
+        // TODO: Combine evidence of window with evidence of value?
+        Justified.byInference(comparison, isWithinWindow, NonEmptyList.of(value, window))
```
jeffmay: I think I can remove the TODO, since the evidence can be merged by a justified interpreter.

core-v1/src/main/scala/algebra/Expr.scala
```diff
@@ -27,15 +30,19 @@ sealed abstract class Expr[-I, +O : OP, OP[_]](val name: String) {
...
     *
     * @note As a top-level [[Expr]] method, this is not very useful, as you cannot invoke it, however, you can
     *       print or inspect it for information at runtime; and placing it here prevents me from forgetting to
-    *       add this to subclasses. The types are refined when attaching the debug hooks using the [[DebugArgs]]
-    *       compiler trick.
+    *       add this to subclasses.
+    *
+    * @see [[Debugging]] for more details on how type checking is handled.
+    *
+    * TODO: Should this be made package private?
```
jeffmay: I think I probably should make the debugging field more private. It would be a confusing interface if you aren't working from within the library.

core-v1/src/main/scala/algebra/Expr.scala
```diff
@@ -256,6 +256,7 @@ object Expr {
     override def withDebugging(debugging: Debugging[Any, Any]): Const[O, OP] = copy(debugging = debugging)
   }
 
+  // TODO: Rename to CombineWith or BinaryOperation to fit better with Zip2With?
   /**
     * Zips the output of the [[leftExpr]] and [[rightExpr]] (executed in parallel, if supported) and passes
     * the tuple as arguments to the [[operation]] that produces a single output value.
```
jeffmay, line 292: `Expr.Combine` is a bit vague. I got it from `Monoid.combine(a, b)`, but it isn't really even the same signature. I should probably do this before v1.0.0, but it might be a separate PR to avoid too much churn.

core-v1/src/main/scala/algebra/Expr.scala
```diff
@@ -296,6 +297,19 @@ object Expr {
...
+  //       Or should we just use the Select() expression?
+  //       ... turns out we need to handle the state and not just the value for certain operations
+  final case class CustomFunction[-I, +O : OP, OP[_]](
+    functionName: String,
+    function: I => O,
+    debugging: Debugging[I, Any] = NoDebugging,
+  ) extends Expr[I, O, OP]("customFunction") {
+    override def visit[G[-_, +_]](v: Visitor[G, OP]): G[I, O] = v.visitCustomFunction(this)
+    override def withDebugging(debugging: Debugging[Any, Any]): Expr[I, O, OP] = copy(debugging = debugging)
+  }
```
jeffmay: I really don't like this hacky escape hatch, but it has proven to be very useful. I think we'll need something like this for the user of the library, no matter what. I think having good metadata here would alleviate my concerns.

I definitely need to document this class though.

core-v1/src/main/scala/dsl/JustifiedBuildExprDsl.scala
```diff
@@ -0,0 +1,119 @@
...
+  override protected implicit final def wrapConst: WrapConst[Justified] = WrapConst.justified
+
+  override protected def wrapElement[C[_], A](
+    outer: Justified[C[A]],
+    element: A,
+  ): Justified[A] =
+    Justified.byInference("elementOf", element, NonEmptyList.of(outer))
+
+  // TODO: Should this be visible outside this trait?
+  protected def dontShortCircuit: Boolean = false
```
jeffmay, line 29: This is a question for future Jeff... I don't know whether short-circuiting will be useful for performance (and thus should be configurable per-DSL), or whether it will not be worth its weight.

core-v1/src/main/scala/dsl/JustifiedBuildExprDsl.scala
```diff
@@ -0,0 +1,119 @@
...
+          combineFalse = NonEmptyList
+            .fromList(_)
+            .map { justified =>
+              Justified.byInference("exists", false, justified)
+            }
+            .getOrElse {
+              // exists in an empty collection is false
+              // TODO: Should I put a reason instead of just a const?
+              //       Maybe I should pass the original value in these functions?
+              Justified.byConst(false)
```
jeffmay, line 75: I think `Justified.byConst` should probably take a reason string. It can default to where the const is defined in code or a constant string like `"source code"` or something

core-v1/src/main/scala/lens/VariantLens.scala
```diff
@@ -0,0 +1,269 @@
...
+import shapeless.ops.hlist
+import shapeless.{Generic, HList}
+
+import scala.collection.{Factory, MapView, View}
+
+// TODO: Rename to NamedLens after old algebra removed
+/**
+  * @see [[NamedLens]]
+  */
+object VariantLens extends VariantLensLowPriorityImplicits {
```
jeffmay: I can't do this until after Kerna migrates to the new algebra.

core-v1/src/main/scala/lens/VariantLens.scala
```diff
@@ -0,0 +1,269 @@
...
+      * Same as [[toMap]], but creates a lazy [[MapView]] instead.
+      */
+    def toMapView: VariantLens[A, MapView[K, V]] = lens.copy(
+      get = lens.get.andThen(Map.from(_).view),
+    )
+  }
+
+  // TODO: Make this more generally useful and convert back to a NamedLens implicitly
+  // I think this can be folded into VariantLens now that it utilizes variance
+  final class AsIterableBuilder[A, C[x] <: IterableOnce[x], E](private val lens: VariantLens[A, C[E]]) extends AnyVal {
```
jeffmay, line 107: I think this TODO is possible, but I don't have time to experiment with this for the first milestone release.

core-v1/src/main/scala/lens/VariantLens.scala
```diff
@@ -0,0 +1,269 @@
...
+    */
+  def asHList[C >: B](implicit gen: Generic[C]): VariantLens[A, gen.Repr] = copy(get = this.get.andThen(gen.to(_)))
+
+  // TODO: Are these runtime down casts needed? Why not just use .asInstanceOf (with a safe wrapper)
+
+//  def asIterable[V](implicit ev: B <:< IterableOnce[V]): VariantLens.AsIterableBuilder[A, IterableOnce, V] =
+//    new VariantLens.AsIterableBuilder(this.copy(get = this.get.andThen(ev)))
+//
+//  def asMap[K, V](implicit ev: B <:< IterableOnce[(K, V)]): VariantLens.AsMapBuilder[A, IterableOnce, K, V] =
+//    new VariantLens.AsMapBuilder(this.copy(get = this.get.andThen(ev)))
```
jeffmay: We need some unit tests in this code to determine whether I can remove this TODO.

core-v1/src/main/scala/lens/VariantLens.scala
```diff
@@ -0,0 +1,269 @@
...
+    */
+  def andThen[C](lens: VariantLens[B, C]): VariantLens[A, C] = {
+    copy(
+      path = this.path ++ lens.path,
+      get = get.andThen(lens.get),
+    )
+  }
+
+  // Helpful for building a tuple with the lens itself. Identical to identity, but you don't have to specify the type.
+  def self: VariantLens[A, B] = this
```
jeffmay: This seems strange. If I'm not using it, I should remove it.

core-v1/src/main/scala/lens/VariantLens.scala
```diff
@@ -0,0 +1,269 @@
...
+    *
+    * Typically, you would not call this directly, but instead by calling the [[NamedLens.Selector.select]]
+    * extension method on this lens. The `.select` method uses macros to make sure that the `name` param is
+    * set properly.
+    *
+    * @param name the name of the field
+    * @param getter a function to extract the value of this field from the object
+    * TODO: Use HList to make this more safe, instead of relying on the macro to use this properly.
+    */
+  def field[C](
```
jeffmay: I doubt I'll ever get around to this TODO. I should probably remove it. Just being careful and using the macro seems to be pretty safe. I doubt adding shapeless will do anything other than introduce difficulty without much benefit.

core-v1/src/test/scala/SimpleArithmeticSpec.scala
```diff
@@ -0,0 +1,46 @@
...
+  test("Duration + Instant") {
+    val now = Instant.now()
+    val duration = Duration.ofMinutes(5)
+    val expr = duration.const + now.const
+    val observed = expr.run()
+    val expected = now.plus(duration)
+    assertEquals(observed, expected)
+  }
+
+  // TODO: LocalDate + Period
```
jeffmay, line 45: I should do this TODO.

core-v1/src/test/scala/SimpleJustifiedArithmeticSpec.scala
```diff
@@ -0,0 +1,61 @@
...
+    val observed = expr.run()
+    val expected = Justified.byInference(
+      "add",
+      now.plus(duration),
+      NonEmptyList.of(Justified.byConst(duration), Justified.byConst(now)),
+    )
+    assertEquals(observed, expected)
+  }
+
+  // TODO: LocalDate + Period
```
jeffmay, line 60: Do this TODO

core-v1/src/main/scala/algebra/ExprResult.scala
```diff
@@ -69,23 +70,31 @@ object ExprResult {
...
+    def visitCustomFunction[I, O : OP](result: CustomFunction[PO, I, O, OP]): I ~> O
 
-    def visitForAll[C[_] : Foldable, E](
-      result: ForAll[PO, C, E, OP],
-    )(implicit
-      opO: OP[Boolean],
-    ): C[E] ~> Boolean
+    def visitExists[C[_] : Foldable, A, B : OP](result: Exists[PO, C, A, B, OP]): C[A] ~> B
+
+    def visitForAll[C[_] : Foldable, A, B : OP](result: ForAll[PO, C, A, B, OP]): C[A] ~> B
```
jeffmay: I should make the same change for `~:>` here as well... maybe call it `~>:` to indicate that the remaining work is applied _after_ the result value has been computed?

core-v1/src/main/scala/algebra/CompareWrapped.scala
jeffmay: Also, this definition is trivial... I don't think I need to do anything here.

circe-v1/src/main/scala/dsl/circe/CirceVaporsEncoders.scala
```diff
@@ -13,23 +13,35 @@ import sourcecode.{File, Line}
...
-      JsonObject(
-        "input" -> state.input.asJson,
-        "output" -> state.output.asJson,
-      )
-    }
+  // TODO: Define this properly
+  implicit val encodeEvidence: Encoder[Evidence] = Encoder.instance { evidence =>
+//    evidence.factSet.map(_.asJson)
+    ???
+  }
```
jeffmay: I don't have time to figure this one out for the first milestone, but I want to solve this before v1.0.0. I think we need to pack the encoder into the `Justified` type (maybe adding an `OP` there?) Not sure, but when I tried this, the changes rippled throughout the code in a way that I wasn't comfortable with tackling just yet.

core-v1/src/main/scala/algebra/CompareWrapped.scala
jeffmay: I removed the TODO, but I think I should really consider implementing a `MonoidK[Justified]` and using that here.

core-v1/src/main/scala/data/ExtractValue.scala
```diff
@@ -0,0 +1,26 @@
+package com.rallyhealth.vapors.v1
+
+package data
+
+/**
+  * Extracts a value from some starting value.
+  */
+trait ExtractValue[-T, +V] {
```
jeffmay: I should probably avoid variance here... most libraries that define typeclasses like this are invariant.

core-v1/src/main/scala/data/Justified.scala
```diff
@@ -2,34 +2,139 @@ package com.rallyhealth.vapors.v1
...
+    reason: String,
+    value: V,
+    sources: NonEmptyList[Justified[Any]],
+  ): Justified[V] = ByInference(reason, value, sources)
+
+  final case class ByInference[V](
+    reason: String,
     value: V,
-    evidence: Evidence,
+    sources: NonEmptyList[Justified[Any]], // TODO: Should this be a NonEmptySet? How to sort it?
```
jeffmay: It seems like most operations have a specific order (like left & right, input & output, etc). Since we can't rely on type information, I don't think sorting is a good idea. I should remove this TODO

core-v1/src/main/scala/engine/StandardEngine.scala
```diff
@@ -126,22 +174,49 @@ object StandardEngine {
     ): PO <:< I => ExprResult[PO, I, Boolean, OP] = { implicit evPOisI =>
       val left = expr.leftExpr.visit(this)(implicitly)
       val right = expr.rightExpr.visit(this)(implicitly)
-      val output = left.state.output || right.state.output
+      val leftIsTrue = left.state.output
+      val rightIsTrue = right.state.output
+      val output = leftIsTrue || rightIsTrue
       // TODO: Do justification here
```
jeffmay, line 183: Justification has moved to a `Justified` wrapper type, so it isn't part of the engine anymore. I should remove this TODO.

circe-v1/src/main/scala/dsl/circe/CirceVaporsEncoders.scala
jeffmay: I can just serialize this as a string for now.

core-v1/src/main/scala/algebra/CompareWrapped.scala
jeffmay: Actually, I can't do this because `Window[V]` and `V` are not the same type, so they cannot be combined into a single result value of the same type. Nor is this `Semigroupal` because the return type `Boolean` is not a product of `V` and `Window[V]`

core-v1/src/main/scala/algebra/ExprResult.scala
```diff
@@ -164,41 +172,87 @@ object ExprResult {
...
+  final case class ValuesOfType[+PO, T, +O, OP[_]](
+    expr: Expr.ValuesOfType[T, O, OP],
+    state: ExprState[PO, Seq[O]],
+  )(implicit
+    opTs: OP[Seq[O]],
+  ) extends ExprResult[PO, Any, Seq[O], OP] {
+    override def visit[G[-_, +_]](v: Visitor[PO, G, OP]): G[Any, Seq[O]] = v.visitValuesOfType(this)
+  }
+
+  final case class WithinWindow[+PO, -I, +V, F[+_], OP[_]](
```
jeffmay: I should document this expression node type.

core-v1/src/main/scala/algebra/ExprResult.scala
```diff
@@ -164,41 +172,87 @@ object ExprResult {
...
   )(implicit
-    opO: OP[Boolean],
-  ) extends ExprResult[PO, C[E], Boolean, OP] {
-    override def visit[G[-_, +_]](v: Visitor[PO, G, OP]): G[C[E], Boolean] = v.visitForAll(this)
+    opO: OP[C[B]],
+  ) extends ExprResult[PO, C[A], C[B], OP] {
+    override def visit[G[-_, +_]](v: Visitor[PO, G, OP]): G[C[A], C[B]] = v.visitMapEvery(this)
+  }
+
+  final case class Not[+PO, -I, +O : Negation : OP, OP[_]](
```
jeffmay, line 224: Document

core-v1/src/main/scala/algebra/ExprResult.scala
```diff
@@ -164,41 +172,87 @@ object ExprResult {
...
+    expr: Expr.Not[I, O, OP],
+    state: ExprState[PO, O],
+    inputResult: ExprResult[PO, I, O, OP],
+  ) extends ExprResult[PO, I, O, OP] {
+    override def visit[G[-_, +_]](v: Visitor[PO, G, OP]): G[I, O] = v.visitNot(this)
+  }
+
+  final case class Select[+PO, -I, +O : OP, OP[_]](
```
jeffmay, line 235: Document

core-v1/src/main/scala/algebra/WrapConst.scala
```diff
@@ -0,0 +1,27 @@
+package com.rallyhealth.vapors.v1
+
+package algebra
+
+import data.Justified
+
+import cats.Id
+
+trait WrapConst[F[_]] {
```
jeffmay: Document

core-v1/src/main/scala/dsl/DslTypes.scala
```diff
@@ -0,0 +1,28 @@
...
+
+  final type OPW[O] = OP[W[O]]
+
+  type W[+V]
+
+  final type FOP[F[_], O] = F[OP[O]]
+
+  final type ~:>[-I, +O] = Expr[I, O, OP]
+
+  final type =~:>[I, +O] = Expr.Identity[I, OP] => Expr[I, O, OP]
```
jeffmay: I should create another non-symbolic alias for this.

core-v1/src/main/scala/dsl/DslTypes.scala
```diff
@@ -0,0 +1,28 @@
+package com.rallyhealth.vapors.v1
+
+package dsl
+
+import algebra.Expr
+
+trait DslTypes extends Any {
```
jeffmay, line 10: I should document all the types in here.

core-v1/src/main/scala/dsl/JustifiedBuildExprDsl.scala
jeffmay: This is probably worth doing before merging.

core-v1/src/main/scala/algebra/Expr.scala
```diff
@@ -135,63 +132,72 @@ object Expr {
...
-    def visitValuesOfType[T](expr: ValuesOfType[T, OP])(implicit opTs: OP[Seq[T]]): Any ~> Seq[T]
+    def visitWithinWindow[I, V : OP, F[+_]](
+      expr: WithinWindow[I, V, F, OP],
+    )(implicit
+      comparison: WindowComparable[F, OP],
+      opB: OP[F[Boolean]],
+    ): I ~:> F[Boolean]
   }
 
   final case class And[-I, OP[_]](
```
jeffmay, line 215: Document

core-v1/src/main/scala/algebra/Expr.scala
```diff
@@ -135,63 +132,72 @@ object Expr {
...
   )(implicit
     opO: OP[Boolean],
   ) extends Expr[I, Boolean, OP]("and") {
-    override def withDebugging(debugging: Debugging[Any, Any]): Expr[I, Boolean, OP] = copy(debugging = debugging)
     override def visit[G[-_, +_]](v: Visitor[G, OP]): G[I, Boolean] = v.visitAnd(this)
+    override private[v1] def withDebugging(debugging: Debugging[Any, Any]): Expr[I, Boolean, OP] =
+      copy(debugging = debugging)
   }
 
   final case class Or[-I, OP[_]](
```
jeffmay, line 235: Document

core-v1/src/main/scala/dsl/standard.scala
```diff
@@ -4,11 +4,16 @@ package dsl
 
 import debug.HasSourceCodeInfo
 
-object standard extends BuildExprDsl with StandardRunDsl {
+case object standard extends UnwrappedBuildExprDsl with StandardRunDsl {
```
jeffmay: TBH, I don't really like the name "standard" here... I don't think having every aspect of the computation captured in a mirror `ExprResult` tree is really "standard". It's more like "lossless" or maybe "mirrored" would be a better name... Either way, this is outside the scope of this PR. I would need to rename a lot of stuff all at once, so this would be a breaking pre-v1 change after I make all the other changes.

vinit-garg (approved):

core-v1/src/main/scala/data/Justified.scala
jeffmay: I left more detailed TODOs describing this comment for future development

core-v1/src/test/scala/SimpleArithmeticSpec.scala
jeffmay: I do this in a later PR: https://github.com/jeffmay/vapors/pull/76

core-v1/src/test/scala/SimpleJustifiedArithmeticSpec.scala
jeffmay: I do this in https://github.com/jeffmay/vapors/pull/76

darmbrus (approved):
The extra documentation on types really helps, thanks for adding that!

vinit-garg (approved):

darmbrus (approved):

PR #69, opened 2021-11-16
Implement a simplified algebra using variance that supports debugging and a higher-kind result parameter (part 1 of 3+)

jeffmay:
- New algebra Expr, ExprResult, and function interpreter compile
  - Support `Expr.ValuesOfType`, `Expr.Const`, `Expr.Identity`
  - Add `Expr.AndThen` for passing the output to the input of another expr
  - Break double operation `Expr.Exists` into a single operation and add an Apply operation
  - Support logical `Expr.Or` and `Expr.And`
- Support serializing expression results as Json with Circe
- Support `+` operation
  - Fix edge case with `+` method and `Predef.any2String` implicit conversion
- Support attaching a debugger
  - Create DebugArgs for type-level calculation of debug parameters
  - Better types for `Expr.Combine` and fix type inference in IntelliJ
  - Document the new pattern for debugging
- Cleanup build
  - Separate core-v1 dependencies
  - Separate circe project dependencies
  - Use package prefixes for new projects
  - Upgrade sbt and plugins
- Create dsl objects for importing appropriate methods, extension methods, type bounds for output parameters (`OP[_]`), and result wrappers (`W[_]`)
  - Create separate DSLs for simple and standard (for re-interpretation) results


core-v1/src/main/scala/engine/SimpleEngine.scala
```diff
@@ -0,0 +1,83 @@
...
+
+    override def visitAndThen[II, IO : OP, OI, OO : OP](
+      expr: Expr.AndThen[II, IO, OI, OO, OP],
+    )(implicit
+      evBI: IO <:< OI,
+    ): II => OO = { ii =>
+      val io = expr.inputExpr.visit(this)(ii)
+      val oo = expr.outputExpr.visit(this)(io)
+      oo
+    }
```
jeffmay, line 36: The `SimpleEngine` does not call the attached debugger, but the `StandardEngine` does... I should either remove debugging or make it call it here.

jeffmay: I make this change in a later commit that is bound up with adding unit tests for this very change. These unit tests depend on a LOT of other changes that I don't want to bring in... so I think I'm gonna skip this comment for now. It will be in the [next PR](https://github.com/jeffmay/vapors/pull/68/files#diff-0eb9ebcb364cc98bd88c65e0bf921f426b8e746c6d5a0ad92fbc203160620562R45)

core-v1/src/main/scala/engine/StandardEngine.scala
```diff
@@ -0,0 +1,147 @@
...
+
+    override def visitAnd[I](
+      expr: Expr.And[I, OP],
+    )(implicit
+      opO: OP[Boolean],
+    ): PO <:< I => ExprResult[PO, I, Boolean, OP] = { implicit evPOisI =>
+      val left = expr.leftExpr.visit(this)(implicitly)
+      val right = expr.rightExpr.visit(this)(implicitly)
+      val output = left.state.output || right.state.output
+      // TODO: Do justification here
```
jeffmay, line 40: This is done in the next PR.

core-v1/src/main/scala/math/Add.scala
```diff
@@ -0,0 +1,67 @@
...
+
+  implicit val addPeriodToLocalDate: Add.Aux[LocalDate, Period, LocalDate] = Add(_.plus(_))
+
+  implicit val addLocalDateToPeriod: Add.Aux[Period, LocalDate, LocalDate] = addPeriodToLocalDate.swapArgs
+}
+
+trait LowPriorityJavaTimeAddImplicits {
+
+  // TODO: Add generic temporal addition operations
+}
```
jeffmay: These generic temporal operations would not be type-safe as they could throw exceptions with the wrong duration or period type. I'm kinda hesitant to add them without returning some kind of option type, but I don't think we ever use the generic types... Maybe I should just remove this for now and let this be a discussion for the future when someone needs it.

core-v1/src/main/scala/data/ExprState.scala
```diff
@@ -0,0 +1,88 @@
+package com.rallyhealth.vapors.v1
+
+package data
+
+// TODO: Figure out how to incorporate justification
```
jeffmay, line 5: This is done in the [next PR](https://github.com/jeffmay/vapors/pull/68)

core/src/main/scala/vapors/data/FactTypeSet.scala
```diff
@@ -6,6 +6,9 @@ import cats.data.{NonEmptyList, NonEmptyMap, NonEmptySet}
 
 import scala.collection.immutable.{SortedMap, SortedSet}
 
+// TODO: Is there a better way to do this?
+//       Maybe we can use something like the Has[_] data type to combine multiple fact types?
+
```
jeffmay, line 11: As far as I can tell, doing a heterogeneous set of fact types would be a pretty big feat. I don't think I can tackle this before launching 1.0.0-M1

circe-v1/src/main/scala/dsl/circe/HasEncoder.scala
```diff
@@ -0,0 +1,20 @@
+package com.rallyhealth.vapors.v1
+
+package dsl.circe
+
+import io.circe.Encoder
+
+// TODO: Figure out how to mix and match DSLs
```
jeffmay: I'm not sure what this means anymore, lol... I think I either did this in this PR OR I am referring to the ability to combine multiple higher-kinded traits using the `with` operator (which I don't think works, but maybe I should try it out). Either way, I should remove this TODO or make it more clear.

jeffmay: I determined that this is not an issue. I am able to use type bounds to constrain subclasses, and then in the specific DSL class, I just pick a single type for which I can combine everything. DSLs are not created on the fly very often, if at all, so this isn't really an issue. I removed the TODO.

core-v1/src/main/scala/algebra/Expr.scala
```diff
@@ -0,0 +1,355 @@
...
+    * This is the magic that allows you to traverse the entire expression recursively. You need a
+    * visitor type that is contravariant for the final input type and covariant for the final output type.
+    *
+    * The visitor is also parameterized on the "output parameter" `OP[_]`, which is supplied at every node
+    * within the given expression tree. This can be useful information, such as how to serialize the output
+    * of every subexpression, or information about the source code position provided by the compiler, etc.
+    *
+    * Every [[Expr]] subclass will call the associated `visitExpr` method and pass the required constraints.
+    *
+    * @tparam ~> an infix type alias for the higher-kinded result type of this visitor
```
paul-a-kennedy-rally, line 138:
```suggestion
    * @tparam ~> an infix type alias for the higher-kinded result type of this visitor (I ~> O is equivalent to ~>[I,O])
```

core-v1/src/main/scala/algebra/ExprResult.scala
```diff
@@ -0,0 +1,204 @@
...
+
+    def visitIdentity[I, O : OP](result: Identity[PO, I, O, OP])(implicit ev: I <:< O): I ~> O
+
+    def visitOr[I](result: Or[PO, I, OP])(implicit opO: OP[Boolean]): I ~> Boolean
+
+    def visitValuesOfType[T](result: ValuesOfType[PO, T, OP])(implicit opTs: OP[Seq[T]]): Any ~> Seq[T]
+  }
+
+  /**
+    * The result of running [[Expr.Or]]
```
paul-a-kennedy-rally:
```suggestion
    * The result of running [[Expr.And]]
```

core-v1/src/main/scala/dsl/RunExprDsl.scala
```diff
@@ -0,0 +1,52 @@
+package com.rallyhealth.vapors.v1
+
+package dsl
+
+import algebra.CombineHolder
+import data.{ExprState, FactTable}
+
+trait RunExprDsl extends CommonDsl {
```
paul-a-kennedy-rally, line 8: Is it possible to add some tparams?  This is a little naked and if you didn't know to go back to Expr.scala, then what might you do?

core-v1/src/main/scala/data/package.scala
```diff
@@ -0,0 +1,16 @@
+package com.rallyhealth.vapors.v1
+
+package object data {
+
+  /**
+    * An ordered set of untyped [[Fact]]s.
+    */
+  final type FactSet = Set[Fact]
```
paul-a-kennedy-rally, line 8: How is this `Set` ordered?

How about `UntypedFactSet`?

core-v1/src/main/scala/data/Justified.scala
```diff
@@ -0,0 +1,35 @@
+package com.rallyhealth.vapors.v1
+
+package data
+
+sealed abstract class Justified[+V] {
```
paul-a-kennedy-rally, line 5: Since this isn't used anywhere that my IDE can detect, I think that we need some more docs on this.

paul-a-kennedy-rally (approved):
I'm curious to see how this evolves.
 * There are some names that I think should be changed--and now is the time to do this.
 * There some docs that can be fixed -- I've added suggestions where it's easy to do so.
 * There are things with no use in the PR and no docs -- I'd like something in place to get an idea of what's going on (tests or docs or both).

I'll approve with the understanding that you'll need another approval for some of the easy changes I've suggested.

darmbrus (approved):

core-v1/src/main/scala/data/Justified.scala
jeffmay: This is used in a future PR. I think I add docs on that PR as well. Sorry for the messy scattershot on implementing this.

core-v1/src/main/scala/algebra/Expr.scala
jeffmay: Infix type notation is a feature of the Scala language, so I'm not sure how I feel about documenting it here. I am wondering about overloading the `~>` symbol as it is used by cats and shapeless for natural transformations (which is not what this is). I'm thinking maybe `~:>` would be better? Two dots to represent the fact that the operation is still pending (kinda like a sideways elipsis `...`) parameterization and the arrow to imply direction from input to output. I could also just go nuts and use a unicode symbol with `⋯>`. The implementer of this trait would not need to use the symbol, as it will get filled in with their custom parameter that they pass.

paul-a-kennedy-rally: _Infix type notation is a feature of the Scala language, so I'm not sure how I feel about documenting it here._ 

I suppose it really depends upon your motivation.  I spoke up about it because I didn't know it.  I usually use one person asking a question as an indication that more people have the same question (or new team members will have) and so clarification is at least necessary.

As to `~>` of `~:>` or just using a letter like we did in the discussion, I am **strongly** in favour of making the code more readable (and less like, say, Perl).

mlinenweber (approved):

core-v1/src/main/scala/dsl/RunExprDsl.scala
jeffmay: Not sure I understand what you mean by this.. What type parameters would I add? What do you have to lookup in Expr.scala?

paul-a-kennedy-rally: It would be nice if there were local definitions for all the type parameters being used here or just better global documentation.  There is neither.

Said another way, if I'm trying to work in the code, I cannot just look at this file and have any sort of link back to what I am assuming is a naming convention.

PR #72, opened 2021-11-17
Add SimpleCachingEngine and create initial benchmark project for comparing performance

jeffmay:
Used to create this [spreadsheet](https://docs.google.com/spreadsheets/d/1sZxsFUm222eXQYcKR_-TAwglTHOmhz5lf17JbiQa6gA/edit?usp=sharing)

- Implement SimpleCachingEngine and add performance tests
  - Add SimpleCachingEngine that utilizes memoization of sub-expressions
  - Create new bench project for performance testing
  - Use a matrix of params for number of expressions in benchmark test
  - Use higher-kinded cache wrapper type for CommonEngine
  - Document the CommonEngine.{exists, forall} methods
  - Use NonEmptySeq instead of NonEmptyList to improve performance of append operation and hide details
- Remove Justified.byInference methods that accept NonEmptyList / NonEmptyVector

### DIFF
--- a/core-v1/src/main/scala/debug/DebugArgs.scala
+++ b/core-v1/src/main/scala/debug/DebugArgs.scala
@@ -3,10 +3,9 @@ package com.rallyhealth.vapors.v1
 package debug
 
 import algebra.Expr
+import cats.data.{NonEmptySeq, NonEmptyVector}
 import data.{ExprState, Window}
 import lens.VariantLens
-
-import cats.data.{NonEmptyList, NonEmptyVector}
 import izumi.reflect.Tag
 
 import scala.reflect.ClassTag
@@ -183,9 +182,9 @@ object DebugArgs {
     A,
     B,
     OP[_],
-  ]: Aux[Expr.Exists[C, A, B, OP], OP, (C[A], Either[List[B], NonEmptyList[B]]), B] =
+  ]: Aux[Expr.Exists[C, A, B, OP], OP, (C[A], Either[Seq[B], NonEmptySeq[B]]), B] =
     new DebugArgs[Expr.Exists[C, A, B, OP], OP] {
-      override type In = (C[A], Either[List[B], NonEmptyList[B]])
+      override type In = (C[A], Either[Seq[B], NonEmptySeq[B]])
       override type Out = B
     }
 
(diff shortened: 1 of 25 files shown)


core-v1/src/main/scala/engine/SimpleCachingEngine.scala
```diff
@@ -0,0 +1,319 @@
...
+    )(implicit
+      debugArgs: DebugArgs[E, OP],
+    ): InvokeAndReturn[E, OP, debugArgs.In, debugArgs.Out] =
+      new InvokeAndReturn(DebugArgs[OP].of(expr)(debugArgs))
+
+    // TODO: Use Hash type bounds on I?
+    protected def memoize[I, O](
+      expr: Expr[I, O, OP],
+      input: I,
+    )(
```
jeffmay, line 119: This would be the safest option, but it might make the code more complicated.

core-v1/src/main/scala/engine/SimpleCachingEngine.scala
```diff
@@ -0,0 +1,319 @@
...
+    }
+
+    override def visitForAll[C[_] : Foldable, A, B : AsBoolean : OP](
+      expr: Expr.ForAll[C, A, B, OP],
+    ): C[A] => CachedResult[B] = memoize(expr, _) { ca =>
+      // TODO: Fix this to match exists
+      val isMatchingResult = expr.conditionExpr.visit(this)
+      val (results, o) = visitForAllCommon(expr, ca)(isMatchingResult.andThen(_.value))
+      debugging(expr).invokeAndReturn(state((ca, results), cached(o)))
+    }
```
jeffmay: BLKR: Do this TODO

core-v1/src/main/scala/debug/DebugArgs.scala
```diff
@@ -183,9 +183,9 @@ object DebugArgs {
     A,
     B,
     OP[_],
-  ]: Aux[Expr.Exists[C, A, B, OP], OP, (C[A], Either[List[B], NonEmptyList[B]]), B] =
+  ]: Aux[Expr.Exists[C, A, B, OP], OP, (C[A], Either[Vector[B], NonEmptyVector[B]]), B] =
```
jeffmay: I think I should use `Seq` and `NonEmptySeq` here to hide implementation details, in case another data type would be more performant...

core-v1/src/main/scala/engine/SimpleCachingEngine.scala
jeffmay: I could potentially store the hash from the previous output (assuming it isn't `ExprState.Nothing`)... I don't think this is really necessary yet though. I might just do a little experiment to prove that the current interface could handle this change in the future.

vinit-garg (approved):

rajatkumar-agarwal (approved):
approved

vinit-garg (approved):

PR #74, opened 2021-11-23
Add Expr.Filter

jeffmay:
- Cleanup some RunDsl types and documentation
- Add Expr.Filter, DSL, and unit tests

### DIFF
--- a/core-v1/src/main/scala/engine/SimpleCachingEngine.scala
+++ b/core-v1/src/main/scala/engine/SimpleCachingEngine.scala
@@ -4,12 +4,12 @@ package engine
 
 import algebra.{EqualComparable, Expr, WindowComparable}
 import cats.data.NonEmptyVector
+import cats.{Eval, Foldable, Functor, FunctorFilter}
 import data.ExtractValue.AsBoolean
-import data.{ExprState, Extract, FactTable, Window}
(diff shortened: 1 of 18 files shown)


core-v1/src/main/scala/engine/SimpleCachingEngine.scala
```diff
@@ -186,6 +186,17 @@ object SimpleCachingEngine {
       debugging(expr).invokeAndReturn(state((ca, results), cached(o, finalState.cacheState)))
     }
 
+    override def visitFilter[C[_] : FunctorFilter, A, B : ExtractValue.AsBoolean : OP](
+      expr: Expr.Filter[C, A, B, OP],
+    )(implicit
+      opO: OP[C[A]],
+    ): C[A] => CachedResult[C[A]] = memoize(expr, _) { input =>
+      // TODO: Is it possible to use cache between elements in the functor?
+      val isMatchingResult = expr.conditionExpr.visit(this).andThen(r => ExtractValue.asBoolean(r.value))
```
jeffmay, line 195: This would require `C[_]` to be `Foldable`, which I don't necessarily want to do. This can be improved with a mutable cache if necessary OR we can require every collection type to be foldable (which is a fairly reasonable restriction).

vinit-garg (approved):

rajatkumar-agarwal (approved):

PR #76, opened 2021-11-28
Add math operators

jeffmay:
- Create CustomAssertions
  - add assertEqualsOrNaN
  - add assertMatchingOutcomes
- Add Subtract definitions and unit tests
- Update Add definition to match Subtraction patterns
  - Add more unit tests for different types
- Define Multiply and add unit tests
- Create Divide operation and add unit tests
  - Add divide by zero TODO for future work
- Add Power operation and add unit tests
- Fix math unit tests for different versions of the JDK


core-v1/src/main/scala/dsl/BuildExprDsl.scala
```diff
@@ -28,6 +29,16 @@ trait BuildExprDsl extends DebugExprDsl {
...
 
+  final def pow[I, L, R](
+    leftExpr: I ~:> W[L],
+    rightExpr: I ~:> W[R],
+  )(implicit
+    opR: OP[W[R]],
+    pow: Power[W[L], W[R]],
+  ): CombineHolder[I, W[L], W[L], W[R], W[R], pow.Out, OP] = {
+    (leftExpr ^ rightExpr)(opR, pow)
+  }
```
jeffmay: I should let DSL subclasses define this to avoid unnecessary wrapper types hanging around in the unwrapped case.

core-v1/src/main/scala/math/Subtract.scala
```diff
@@ -0,0 +1,56 @@
...
+object Subtract extends SubtractNumericImplicits with SubtractJavaTimeImplicits {
+  type Aux[L, R, O] = Subtract[L, R] { type Out = O }
+
+  def apply[L, R, O](fn: (L, R) => O): Subtract.Aux[L, R, O] = new Subtract[L, R] {
+    override type Out = O
+    override def subtract(
+      left: L,
+      right: R,
+    ): Out = fn(left, right)
+  }
```
jeffmay: This is not a common use of the `.apply` method. I should rename this to `.instance` or maybe `.define`

core-v1/src/main/scala/algebra/Expr.scala
```diff
@@ -119,7 +119,73 @@ sealed abstract class Expr[-I, +O : OP, OP[_]](val name: String) extends Product
...
+    * @param that the other expression to subtract from the output of this expression
+    * @param sub the type-level definition of how to subtract that type of output from this type of output
+    *
+    * @return a [[CombineHolder]] to allow for type-level calculation of the return type
+    */
+  def -[CI <: I, LI >: O, RI >: RO, RO <: RI : OP](
+    that: Expr[CI, RO, OP],
+  )(implicit
+    sub: Subtract[LI, RI],
+  ): CombineHolder[CI, LI, O, RI, RO, sub.Out, OP] =
```
jeffmay: I should document all of the type parameters for these

core-v1/src/main/scala/algebra/Expr.scala
```diff
@@ -119,7 +119,73 @@ sealed abstract class Expr[-I, +O : OP, OP[_]](val name: String) extends Product
...
+  }
+
+  def ^[CI <: I, LI >: O, RI >: RO, RO <: RI : OP](
+    that: Expr[CI, RO, OP],
+  )(implicit
+    pow: Power[LI, RI],
+  ): CombineHolder[CI, LI, O, RI, RO, pow.Out, OP] = {
+    // can't eta-expand a dependent object function, the (_, _) is required
+    new CombineHolder(this, that, "pow", pow.power(_, _): @nowarn)
   }
```
jeffmay: BLKR: This needs documentation

vinit-garg (approved):

core-v1/src/test/scala/SimpleDivisionSpec.scala
```diff
@@ -0,0 +1,125 @@
...
+  import dsl.simple._
+
+  // TODO: Should this return an Option[Int]? Maybe another method for this?
+  test("Divide by zero") {
+    val expr = 1.const / 0.const
+    val exc = intercept[ArithmeticException] {
+      expr.run()
+    }
+    assert(exc.getMessage contains "zero")
+  }
```
jeffmay, line 16: I'm leaving this for a future change. While I don't think expressions should throw exceptions when interpreted. I don't think using an `Option` interface is great either, as there isn't really a way to recover from a divide by zero error any better than just throwing that exception, and there isn't any way to know apriori that the denominator expression would return a `0`.

darmbrus (approved):

PR #82, opened 2021-12-15
Add concat, wrapAll, flatten, and flatMap DSL methods and unit tests

jeffmay:
- Support `wrapAll` using `Expr.Sequence` to convert a `C[I ~:> A]` into an expression `I ~:> C[A]`
- Support `concat` using `Expr.Sequence` and then `Expr.Flatten` to convert a variable argument list of `C[I ~:> A]` (where `C[_]` is `Foldable`) into an expression `I ~:> Seq[A]`
- Support `.flatMap` and `.flatten` using `Expr.Select` and `Expr.Flatten` to drop to a common super-type of two higher-kinded collections to go from `I ~:> C[D[A]]` into `I ~:> F[F[A]]` and then use `Expr.Flatten` to chain together a final `I ~:> F[A]`
- Clean up some issues with `VariantLens` extension methods


core-v1/src/main/scala/algebra/Expr.scala
```diff
@@ -740,6 +753,16 @@ object Expr {
...
 
+  final case class Flatten[C[_] : FlatMap, A, OP[_]](
+    private[v1] val debugging: Debugging[Nothing, Nothing] = NoDebugging,
+  )(implicit
+    opO: OP[C[A]],
+  ) extends Expr[C[C[A]], C[A], OP]("flatten") {
+    override def visit[G[-_, +_]](v: Visitor[G, OP]): G[C[C[A]], C[A]] = v.visitFlatten(this)
+    override private[v1] def withDebugging(debugging: Debugging[Nothing, Nothing]): Flatten[C, A, OP] =
+      copy(debugging = debugging)
+  }
```
jeffmay, line 789: BLKR: Document this expression node

core-v1/src/main/scala/algebra/Expr.scala
```diff
@@ -821,6 +844,17 @@ object Expr {
...
+  final case class Sequence[C[+_] : Traverse, -I, +O, OP[_]](
+    expressions: C[Expr[I, O, OP]],
+    override private[v1] val debugging: Debugging[Nothing, Nothing] = NoDebugging,
+  )(implicit
+    opO: OP[C[O]],
+  ) extends Expr[I, C[O], OP]("sequence") {
+    override def visit[G[-_, +_]](v: Visitor[G, OP]): G[I, C[O]] = v.visitSequence(this)
+    override private[v1] def withDebugging(debugging: Debugging[Nothing, Nothing]): Sequence[C, I, O, OP] =
+      copy(debugging = debugging)
+  }
```
jeffmay, line 889: BLKR: Document this expression node

core-v1/src/main/scala/dsl/BuildExprDsl.scala
```diff
@@ -32,6 +32,22 @@ trait BuildExprDsl extends DebugExprDsl with WrapArityMethods {
...
+  ): AndThen[I, Seq[Seq[A]], Seq[A]] = {
+    Expr
+      .Sequence(expressions.map {
+        Expr.Select(_, VariantLens.id[C[A]].asIterable.to[Seq], (_: C[A], sa: Seq[A]) => sa)
+      })
+      .andThen(Expr.Flatten())
+  }
+
+  def flatten[I, C[+_] : FlatMap, A](expr: I ~:> C[C[A]])(implicit opCA: OP[C[A]]): AndThen[I, C[C[A]], C[A]] =
+    expr.andThen(Expr.Flatten())
```
jeffmay: I think this should probably be a method on the higher kinded type, rather than a top-level method.

core-v1/src/main/scala/lens/VariantLens.scala
```diff
@@ -144,7 +146,19 @@ object VariantLens extends VariantLensLowPriorityImplicits {
...
-    * Wrap the result of this lens with an [[AsIterableBuilder]] for helper operations on [[IterableOnce]] types.
+    * Wrap the result of this lens with an [[VariantLens.AsIterableBuilder]] for helper operations
+    * on [[IterableOnce]] types.
+    */
+  implicit def asFoldableIterable[A, B, C[_] : Foldable, E](
+    lens: VariantLens[A, B],
+  )(implicit
+    ev: B <:< C[E],
+  ): VariantLens.AsIterableBuilder[A, E] =
+    new VariantLens.AsIterableBuilder(lens.asIterable)
```
jeffmay: This seems to fail when you also have `import cats._` because it will assume the `Foldable[Id]` for collection types... I wonder if there is a way to fix this.

jeffmay (commented):
This needs more unit tests

core-v1/src/main/scala/engine/SimpleCachingEngine.scala
```diff
@@ -30,8 +30,20 @@ object SimpleCachingEngine {
...
+      new Applicative[CachedResult] with Extract[CachedResult] with Foldable[CachedResult] with Functor[CachedResult] {
+
+        override def pure[A](x: A): CachedResult[A] = CachedResult(x, Map.empty)
+
+        override def ap[A, B](ff: CachedResult[A => B])(fa: CachedResult[A]): CachedResult[B] = {
+          val value = ff.value(fa.value)
+          // TODO: What if I just take the cache from fa?
+          val combinedCache = ff.cacheState ++ fa.cacheState
+          CachedResult(value, combinedCache)
+        }
```
jeffmay: I think accepting the cache from only the final result would work in almost every case.

It is possible that the implementation of the given `ff` is incorrect, in that it does not pass the `cacheState` from `ff` into the result of `fa`. This would result in the final result missing some state from the previous cache, but I think the definition of `Applicative` would not be violated in this case.

Only accepting the cache state from `fa` would violate the definition of `Parallel`, as the order of monadic chaining matters for determining what the final cache state is, however, this class is useless in a parallel manner. The whole point of a cache is to remember the result of previous actions, so the order of actions is always relevant and the cache must always be updated between operations. Executing expressions in parallel means that we cannot take advantage of this type of cache. It would be possible to share a cache of atomic pointers in the future, but we would have to reimplement this interpreter and model class anyway.

I think I'm going to just return `fa.cacheState`

vinit-garg (approved):

core-v1/src/main/scala/lens/VariantLens.scala
```diff
@@ -246,8 +254,16 @@ final case class VariantLens[-A, +B](
     * @note this would not be necessary if [[VariantLens]] used appropriate variance, however, variance for lenses
     *       introduces a lot of issues.
     */
-  def as[V](implicit ev: B <:< V): VariantLens[A, V] =
-    this.copy(get = this.get.andThen(ev))
+  def as[V](implicit convert: B => V): VariantLens[A, V] =
+    this.copy(get = this.get.andThen(convert))
```
jeffmay: An implicit function is also know as a "view bound". It had its own syntax `def f[A <% B](a: A)` that was deprecated in Scala 2.11 in favor of just accepting the implicit function argument `def f[A](a: A)(implicit ev: A => B)`.

I'm not sure why I made this change, but I think this is more generically useful, although it could be used for more purposes than just proof of subtyping, which might non-obvious. I should probably revert this if it isn't necessary for some collection operation, since the caller can do this explicitly with the `.copy` method.

darmbrus (approved):

PR #84, opened 2021-12-27
Add Expr.When, DSL methods, and unit tests

jeffmay:
Add Expr.When, DSL methods, and unit tests

### DIFF
--- a/core-v1/src/main/scala/engine/SimpleCachingEngine.scala
+++ b/core-v1/src/main/scala/engine/SimpleCachingEngine.scala
@@ -331,6 +331,32 @@ object SimpleCachingEngine {
       debugging(expr).invokeAndReturn(state(i, cached(o)))
     }
 
+    override def visitWhen[I, B : ExtractValue.AsBoolean, O : OP](expr: Expr.When[I, B, O, OP]): I => CachedResult[O] =
+      memoize(expr, _) { i =>
+        val (cacheState, firstResult) =
+          expr.conditionBranches.zipWithIndex.foldLeft((resultCache, None: Option[(O, Int)])) {
+            case ((cacheState, None), (cb, idx)) =>
+              val condResult = visitWithUpdatedCache(i, cb.whenExpr, cacheState)
+              val condIsMet = ExtractValue.asBoolean(condResult.value)
+              if (condIsMet) {
+                val branchResult = visitWithUpdatedCache(i, cb.thenExpr, condResult.cacheState)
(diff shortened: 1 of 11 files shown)


core-v1/src/main/scala/engine/SimpleCachingEngine.scala
```diff
@@ -331,6 +331,34 @@ object SimpleCachingEngine {
...
+              val condResult = visitWithUpdatedCache(i, cb.whenExpr, cacheState)
+              val condIsMet = ExtractValue.asBoolean(condResult.value)
+              Option
+                .when(condIsMet) {
+                  val branchResult = visitWithUpdatedCache(i, cb.thenExpr, condResult.cacheState)
+                  (branchResult.cacheState, Some(branchResult.value, idx))
+                }
+                .getOrElse {
+                  (condResult.cacheState, None)
+                }
```
jeffmay: I think can use a simple `if` / `else` here

core-v1/src/test/scala/SimpleJustifiedToHListSpec.scala
```diff
@@ -1,139 +0,0 @@
...
-
-import data.Justified
-import example.Person
-import lens.DataPath
-
-import cats.data.NonEmptySeq
-import munit.FunSuite
-import shapeless.HNil
-
-class SimpleJustifiedToHListSpec extends FunSuite {
```
jeffmay: Revert the deletion of this file.

vinit-garg (approved):

core-v1/src/main/scala/engine/StandardEngine.scala
```diff
@@ -269,6 +270,32 @@ object StandardEngine {
       ExprResult.ValuesOfType(expr, finalState)
     }
 
+    override def visitWhen[I, B : ExtractValue.AsBoolean, O : OP](
+      expr: Expr.When[I, B, O, OP],
+    ): PO <:< I => ExprResult[PO, I, O, OP] = { implicit evPOisI =>
+      val initFalseResults: Seq[ExprResult[PO, I, B, OP]] =
+        new Array[ExprResult[PO, I, B, OP]](expr.conditionBranches.length).toSeq
```
jeffmay: This isn't helpful, it creates an Array that must be copied on every write. If I'm gonna do this, I should use a `Vector`, so that the final product is a persistent immutable data structure that doesn't need to be copied again at the end.

darmbrus (approved):

PR #85, opened 2022-01-06
All remaining work for the milestone (M1) release

jeffmay:
- Add min, max, and sum custom functions with unit tests
- Implement Expr.ContainsAny, in / containsAny DSL methods, and unit tests
- Add support for Option.filter
- Fix variance issue with === and =!=
- Rename run implicit to runAny to avoid scope name collision
- Fix bug in when / elif / else expressions
- Fix bug in IsEqual and add unit tests
- Fix name of FoldLeft DebugArgs
- Remove unnecessary SpecificHkExprBuilder type alias
- Support caching between elements of Expr.Sequence
- Improve performance of exists / forall
- Use HMap for SimpleCachingEngine to be more type-safe
- Add CacheRunDsl to run expressions with a cache
  - Factor out common RunExprDsl types
  - Remove use of default arguments
  - Replace simple dsl engine with SimpleCachingEngine


build.sbt
```diff
@@ -4,6 +4,9 @@ name := "vapors-root"
 ThisBuild / organization := "com.rallyhealth"
 ThisBuild / organizationName := "Rally Health"
 
+// publish snapshots until we have a 1.0.0-M1 git tag
+isSnapshot := true
```
jeffmay, line 8: BLKR: Remove this before merging

core-v1/src/main/scala/algebra/Expr.scala
```diff
@@ -746,6 +757,34 @@ object Expr {
...
+    validValuesExpr: Expr[I, Set[W[A]], OP],
+    // TODO: Support short-circuiting by returning a single element?
+    // TODO: Use a Set instead of a List? Seq instead of List? Upcast to Iterable to allow either?
+    foundMatchingElements: (C[W[A]], Set[W[A]], List[W[A]]) => B,
+    override private[v1] val debugging: Debugging[Nothing, Nothing] = NoDebugging,
+  ) extends Expr[I, B, OP]("containsAny") {
+    override def visit[G[-_, +_]](v: Visitor[G, OP]): G[I, B] = v.visitContainsAny(this)
+    override private[v1] def withDebugging(debugging: Debugging[Nothing, Nothing]): ContainsAny[I, W, C, A, B, OP] =
+      copy(debugging = debugging)
+  }
```
jeffmay, line 786: This is the last change in this PR that may require refactoring.

jeffmay: On second thought... I think this should be a M2 improvement.

core-v1/src/main/scala/data/Justified.scala
```diff
@@ -258,6 +258,40 @@ object Justified extends LowPriorityJustifiedImplicits {
...
+      // TODO: Figure out how to avoid the Justified.byConst situation.
+      // One option is to wrap every collection as well as the elements to provide traceability for empty collections
+      val justifiedValidValues = NonEmptySeq
+        .fromSeq(valid.toSeq)
+        .map { nonEmpty =>
+          Justified.byInference("validValues", valid.map(_.value), nonEmpty)
+        }
+        .getOrElse {
+          Justified.byConst(Set())
+        }
```
jeffmay, line 282: I think this can be addressed using a wrapper around a collection of wrapped elements. This would require a significant refactor across the DSL, but I think it should be done pre-1.0.0.

core-v1/src/main/scala/engine/SimpleCachingEngine.scala
```diff
@@ -210,6 +241,27 @@ object SimpleCachingEngine {
...
+      val found = input
+        .foldRight(Eval.now(List.empty[W[A]])) { (wa, acc) =>
+          val a = W.extract(wa)
+          if (valid(a)) acc.map(wa :: _)
+          else acc
+        }
+        .value
+      val o = expr.foundMatchingElements(input, wrappedValidSet, found)
+      debugging(expr).invokeAndReturn(state((i, input, wrappedValidSet, found), cached(o, finalCacheState)))
+    }
```
jeffmay, line 263: These TODOs will be impacted by the refactor of `Expr.ContainsAny`, so I am going to postpone the work on these until those changes are made.

core-v1/src/main/scala/dsl/simple.scala
```diff
@@ -2,12 +2,16 @@ package com.rallyhealth.vapors.v1
...
+  final case object justified extends FullDsl with JustifiedBuildExprDsl with CachingRunDsl with ShowOPDsl {
 
-    final case object withDebugInfo extends FullDsl with JustifiedBuildExprDsl with SimpleRunDsl with DebugSourceInfoDsl
+    final case object withDebugInfo
+      extends FullDsl
+      with JustifiedBuildExprDsl
+      with CachingRunDsl
+      with DebugSourceInfoDsl
   }
 }
```
jeffmay: This change uses caching by default (`dsl.simple` now uses caching). I could alternatively create a `dsl.caching` import to allow the user to decide. I think this might be preferable... maybe `dsl.uncached` / `dsl.caching.mutable` / `dsl.caching.parallel` could be the future state for the DSLs?

vinit-garg (approved):

darmbrus (approved):

PR #86, opened 2022-01-11
Add Expr.FoldLeft, DSL methods and unit tests

jeffmay:


core-v1/src/main/scala/algebra/Expr.scala
```diff
@@ -888,6 +893,20 @@ object Expr {
       copy(debugging = debugging)
   }
 
+  // TODO: Is O needed?
```
jeffmay: BLKR: Remove this TODO and document this class.

core-v1/src/test/scala/SimpleFoldLeftSpec.scala
```diff
@@ -0,0 +1,15 @@
...
+class SimpleFoldLeftSpec extends FunSuite {
+
+  import dsl.simple._
+
+  test("it works") {
+    val expr = Seq(1, 2, 3).const.foldLeft(1.const)(_ + _)
+    val observed = expr.run()
+    val expected = Seq(1, 2, 3).foldLeft(1)(_ + _)
+    assertEquals(observed, expected)
+  }
```
jeffmay: Update test name and add more unit tests

vinit-garg (approved):

darmbrus (approved):

core-v1/src/main/scala/algebra/Expr.scala
```diff
@@ -888,6 +893,40 @@ object Expr {
...
+    *
+    * @param inputExpr an expression that produces a [[Foldable]] output
+    * @param initExpr an expression that produces an initial value for the accumulator (the "left" side)
+    * @param foldExpr an expression that operates on a tuple of the accumulated ("left" side) and the next
+    *                 element of the [[Foldable]] inputExpr result ("right" side) to produce a final output
+    *                 of the accumulated value.
+    *
+    * @tparam C the [[Foldable]] collection type
+    * @tparam A the element type
+    * @tparam B the accumulator type, initial value, and return type
```
paul-a-kennedy-rally: Could we make `E` the element type, and `A` the accumulator type?

paul-a-kennedy-rally: Actually, since we're not limited to one letter type parameters, why not ELT and ACC?

paul-a-kennedy-rally (approved):
I don't want to block this effort and I know that we're in the middle of a series of reviews, but I think making the type parameters a little easier to map would be a helpful addition.

Maybe at the end?

core-v1/src/main/scala/algebra/Expr.scala
jeffmay: I can do that for this one, however, in general, I use `A` for the element type. I used to use `E`, but I think `A` is more common in Scala collections and Cats because it stands for "anything" (i.e. it has no constraints). I also don't think it matters too much because the IDE will replace these parameter names when you look at the type signatures, so it only really matters in local context.

PR #87, opened 2022-01-11
Add .sizeIs / .isEmpty / .head operators including NonEmpty types and improve type inference for constants

jeffmay:


circe-v1/src/test/scala/circe/StandardResultAsJsonSpec.scala
```diff
@@ -14,14 +14,14 @@ class StandardResultAsJsonSpec extends FunSuite {
   import dsl.circe.standard._
 
   test("serialize an expression state without input") {
-    val expr = "World".const
+    val expr = const("World")(constId).const
```
jeffmay: BLKR: Revert

core-v1/src/main/scala/algebra/SizeComparable.scala
```diff
@@ -0,0 +1,73 @@
+package com.rallyhealth.vapors.v1
+
+package algebra
+
+import cats.data.{NonEmptyList, NonEmptySeq, NonEmptyVector}
+
+trait SizeComparable[-C, -N, +B] { outer =>
```
jeffmay, line 15: BLKR: Document. Also, explain why contravariant.

core-v1/src/main/scala/algebra/Expr.scala
```diff
@@ -966,6 +978,18 @@ object Expr {
       copy(debugging = debugging)
   }
 
+  final case class SizeIs[-I, N : ExtractValue.As[Int]#From, B : ExtractValue.As[Boolean]#From : OP, OP[_]](
```
jeffmay: BLKR: Add documentation

core-v1/src/main/scala/data/ExtractValue.scala
```diff
@@ -15,9 +15,17 @@ trait ExtractValue[T, V] extends (T => V) {
 object ExtractValue {
   type AsBoolean[T] = ExtractValue[T, Boolean]
 
+  type As[V] = {
+    type From[T] = ExtractValue[T, V]
+  }
+
+  type From[T] = {
+    type As[V] = ExtractValue[T, V]
+  }
```
jeffmay: Are these being used? If so, I should document them and maybe refactor old code to use them where appropriate. I suppose one advantage of these is that you don't need the kind-projector compiler plugin to add these as context bounds. However, long-term, I think these will be phased out in favor of kind-projector... so why add them now? Are they really used enough to make it worth adding them here?

core-v1/src/main/scala/dsl/ConstOutputType.scala
```diff
@@ -15,16 +15,20 @@ import scala.annotation.implicitNotFound
...
+)
+trait ConstOutputType[W[+_], -O] {
   type Out
 
   def wrapConst(value: O): Out
 }
 
 object ConstOutputType {
-  type Aux[W[+_], O, WO] = ConstOutputType[W, O] { type Out = WO }
+  type Aux[W[+_], -O, WO] = ConstOutputType[W, O] { type Out = WO }
```
jeffmay, line 33: I found this to be helpful when determining the return type of the `.const` operator when working with `Some[A]`, as an implicit for `Option[A]` will be more specific (according to the rules of contravariance) than `Some[A]`, so it allow (and prefer) a `Traverse[Option]` over a `Traverse[Some]` (which does not exist)

jeffmay: Also, I should move this comment about variance to the ScalaDocs

vinit-garg (approved):

mlinenweber (approved):

PR #88, opened 2022-01-12
Add Expr.Define and Expr.UsingDefinitions, DSL, and unit tests

jeffmay:
- Add Expr.Define and Expr.UsingDefinitions, DSL, and unit tests
- Add using syntax for building a dependency on definitions
  - Add UsingFn classes for building expressions from captured function types
  - Add using methods for wrapped and unwrapped cases

### DIFF
--- a/core-v1/src/main/scala/dsl/BuildExprDsl.scala
+++ b/core-v1/src/main/scala/dsl/BuildExprDsl.scala
@@ -12,7 +12,7 @@ import cats.data.{NonEmptySeq, NonEmptyVector}
 import cats.{FlatMap, Foldable, Functor, FunctorFilter, Id, Order, Reducible, Traverse}
 import shapeless.{Generic, HList}
 
-trait BuildExprDsl extends DebugExprDsl with WrapArityMethods {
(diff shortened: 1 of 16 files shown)


core-v1/src/main/scala/dsl/BuildExprDsl.scala
```diff
@@ -53,6 +53,64 @@ trait BuildExprDsl extends DebugExprDsl with WrapArityMethods {
...
+
+  // TODO: Support function builder syntax
+  final def usingDefinitions[I, O](
+    definitions: Expr.Definition[I, OP]*,
+  )(
+    thenExpr: Any ~:> O,
+  )(implicit
+    opO: OP[O],
+  ): Expr.UsingDefinitions[I, O, OP] =
+    Expr.UsingDefinitions(definitions, thenExpr)
```
jeffmay: Remove the TODO, since this is done. Also, maybe deprecate this method for backwards compatibility only?

vinit-garg (approved):

darmbrus (approved):

PR #93, opened 2022-02-04
Add .asCase to VariantLens

jeffmay:
- Add `.asCase` to `VariantLens` using `shapeless.Typeable`

### DIFF
--- a/core-v1/src/main/scala/lens/VariantLens.scala
+++ b/core-v1/src/main/scala/lens/VariantLens.scala
@@ -9,7 +9,7 @@ import cats.data.NonEmptySet
 import cats.kernel.Semigroup
 import cats.{Eval, Foldable, Reducible}
 import shapeless.ops.hlist
-import shapeless.{Generic, HList}
+import shapeless.{Generic, HList, Typeable}
 
 import scala.annotation.nowarn
 import scala.collection.{Factory, View}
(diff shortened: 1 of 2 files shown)


vinit-garg (approved):

core-v1/src/main/scala/lens/VariantLens.scala
```diff
@@ -147,6 +147,15 @@ object VariantLens extends VariantLensLowPriorityImplicits {
...
       )
   }
+
+  implicit final class AsSumTypeBuilder[A, B](private val lens: VariantLens[A, B]) extends AnyVal {
+
+    /**
+      * Attempt to cast the value of super-type [[B]] as sub-type [[V]] (if possible), otherwise None.
+      */
+    def asCase[V <: B : Typeable]: VariantLens[A, Option[V]] =
+      lens.copy(get = lens.get.andThen(Typeable[V].cast))
```
jeffmay, line 157: I think I should handle multiple cases at once, otherwise, you would be required to use `when` / `then` / `else` and unify the results. A `Expr.Match` might be a better option... it could also potentially replace the `When` operator as a more generally useful operator, but each branch has a lens that is applied before any optional condition guard.

mlinenweber (approved):

preetbawa (approved):

jeffmay:
Closing in favor of https://github.com/jeffmay/vapors/pull/95

PR #94, opened 2022-02-10
Add repeat operation and support use case of threading it into zipToShortest

jeffmay:
- Add `Expr.Repeat` to continuously produce the same value indefinitely
- Allow `IterableOnce` for `ExprHList.zipToShortest`
- Use `Seq` for return type of `ZipToShortest` for `IterableOnce`

### DIFF
--- a/core-v1/src/main/scala/algebra/Expr.scala
+++ b/core-v1/src/main/scala/algebra/Expr.scala
@@ -1044,8 +1044,17 @@ object Expr {
       copy(debugging = debugging)
   }
 
+  /**
+    * Creates an [[IterableOnce]] that emits the result of the input expression forever, or up to a given limit.
+    *
+    * @param inputExpr the input expression to repeat
+    * @param recompute whether to recompute the expression on every iteration or just use the first result
+    * @param limit whether to limit the total number of elements produced by the iterable
+    */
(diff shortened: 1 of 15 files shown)


core-v1/src/main/scala/algebra/Expr.scala
```diff
@@ -1039,6 +1044,17 @@ object Expr {
...
+  final case class Repeat[-I, +O, OP[_]](
+    inputExpr: Expr[I, O, OP],
+    override private[v1] val debugging: Debugging[Nothing, Nothing] = NoDebugging,
+  )(implicit
+    opO: OP[IterableOnce[O]],
+  ) extends Expr[I, IterableOnce[O], OP]("repeat") {
+    override def visit[G[-_, +_]](v: Visitor[G, OP]): G[I, IterableOnce[O]] = v.visitRepeat(this)
+    override private[v1] def withDebugging(debugging: Debugging[Nothing, Nothing]): Repeat[I, O, OP] =
+      copy(debugging = debugging)
+  }
```
jeffmay: BLKR: This should have documentation. Also, maybe it would be better to put some kind of limit to avoid an infinite loop.

vinit-garg (approved):

preetbawa (approved):

PR #96, opened 2022-02-17
Update circleci config to publish coverage report to codecov

jeffmay:
Add codecov back using the more secure CircleCI Orb tool (rather than the unsafe bash uploader)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -40,7 +40,7 @@ jobs:
             - v1-dependencies-
 
       # Compile, run tests, and produce coverage reports!
-      - run: sbt clean coverage test coverageReport coverageAggregate shutdown << /dev/null
+      - run: exec 0<&- && sbt clean coverage test coverageReport coverageAggregate shutdown
 
       # Save cached artifacts
       - save_cache:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -40,7 +40,7 @@ jobs:
             - v1-dependencies-
 
       # Compile, run tests, and produce coverage reports!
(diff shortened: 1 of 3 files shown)


vinit-garg (approved):

.circleci/config.yml
```diff
@@ -36,12 +39,14 @@ jobs:
             # fallback to using the latest cache if no exact match is found
             - v1-dependencies-
 
-      - run: cat /dev/null | sbt 'Test / compile'
+      # Compile, run tests, and produce coverage reports!
+      - run: cat /dev/null | sbt clean coverage test coverageReport coverageAggregate shutdown
```
paul-a-kennedy-rally: If you want to close `stdin`, 

```suggestion
      - run: exec 0<&- && sbt clean coverage test coverageReport coverageAggregate shutdown
```

paul-a-kennedy-rally: Or you could redirect in an empty input:
```suggestion
      - run: sbt clean coverage test coverageReport coverageAggregate shutdown << /dev/null
```

Either one of these would stop another shell process from being started.

paul-a-kennedy-rally (commented):
I would fix the shell usage if the build system supports it.  It would start one fewer shell process.

.circleci/config.yml
jeffmay: Neat! I just copied this from the template they posted. I like your solution better, but I would kinda prefer to stick to the template unless there is a good reason to deviate. Is there any advantages to this method?

paul-a-kennedy-rally (approved):
LGTM

PR #101, opened 2022-08-18
FactTable label suggestion.

paul-a-kennedy-rally:
It would be nice to support a debug label for FactTable, for use in testing output.

For example:

    [info]   systolic
    [info]   - tests systolic_lt_120 is false with FactTable(name=BloodPressureSystolic,unit=mmHg,value=120.0)
    [info]   - tests systolic_lt_120 is true with FactTable(name=BloodPressureSystolic,unit=mmHg,value=110.0)
    [info]   - tests systolic_lt_120 is false with FactTable({empty})
    [info]   - tests systolic_ge_120 is false with FactTable(name=BloodPressureSystolic,unit=mmHg,value=110.0)
    [info]   - tests systolic_ge_120 is true with FactTable(name=BloodPressureSystolic,unit=mmHg,value=120.0)
    [info]   - tests systolic_ge_120 is false with FactTable({empty})

By having a debug label, test output can have more salient details for testing without interfering with the toString() output.

This did change the value class semantics, so I don't know whether there is a hidden performance hit.

### DIFF
--- a/core-v1/src/main/scala/data/FactTable.scala
+++ b/core-v1/src/main/scala/data/FactTable.scala
@@ -14,7 +14,7 @@ import scala.collection.immutable.SortedMap
   *
   * @note some expressions can update the fact table for sub-expressions.
   */
-final case class FactTable(factsByName: SortedMap[String, FactSet]) extends AnyVal {
+final case class FactTable(factsByName: SortedMap[String, FactSet], label: String = "{no debug label}") {
 
   def add(fact: Fact): FactTable = addAll(FactSet(fact))
 


jeffmay:
I feel like a more modular approach is to use a tuple (or case class) of `(label: String, facts: FactTable)` instead of putting that into this data type. I understand the use case in testing, but I don't know if it makes sense to put that here. Maybe instead we could make `FactTable` a trait and allow subclassing outside of this repo.

jeffmay:
Closing in favor of https://github.com/jeffmay/vapors/pull/102

PR #107, opened 2022-09-24
Make Fact an open interface and make FactTable an IterableOnce

jeffmay:
* Fact model improvements 
  - Convert Fact to open interface
  - Override TypedFact methods for better type inference
* Extend IterableOnce in FactTable and add unit tests

### DIFF
--- a/core-v1/src/main/scala/data/FactTable.scala
+++ b/core-v1/src/main/scala/data/FactTable.scala
@@ -14,7 +14,7 @@ import scala.collection.immutable.SortedMap
   *
   * @note some expressions can update the fact table for sub-expressions.
   */
-trait FactTable extends Any {
+trait FactTable extends Any with IterableOnce[Fact] {
   protected type Self <: FactTable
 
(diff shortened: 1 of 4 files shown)


core-v1/src/main/scala/data/FactTable.scala
```diff
@@ -14,7 +14,7 @@ import scala.collection.immutable.SortedMap
   *
   * @note some expressions can update the fact table for sub-expressions.
   */
-trait FactTable extends Any {
+trait FactTable extends Any with IterableOnce[Fact] {
```
jeffmay, line 17: This should really be `Iterable`, however, if I want to keep the `universal trait` aspect of this trait, then I can't extend `Iterable` because it doesn't extend `Any`. However, using this as an `Iterable` is not a very common use case and calling `.iterator.toIterable` isn't that complicated so it's good enough.

mingy711 (approved):

darmbrus (approved):

mingy711 (approved):

